Issue #14, opened 2019-01-22
[Proposal] Rules Severities for Marketplace Technical Review


lenaorobei:
### Marketplace Technical Review

PHP CodeSniffer is one of the tools Magento Marketplace uses for extensions verification. Find more details about current approach [here](https://docs.magento.com/marketplace/user_guide/sellers/code-validation.html). 
Severities is the way to determine how strict the rule is. The higher is the severity the stricter the rule is.

⚠️ **Note:** _only severity 10 issues lead to extension rejection._

### Proposed Severities Definition

| Type | Severity | Description |
| --- | --- | --- |
| error | 10 |  Critical code issue. |
| warning | 9 |  Security issue (possible). |
| warning | 8 |  Magento specific code issue. |
| warning | 7 |  General code issue. |
| warning | 6 |  Code style issue (PSR2). |
| warning | 5 |  PHPDoc formatting or commenting issue. |

### Proposed Rules Severities

The Magento standard contains 96 sniffs.

Fixable column identifies whether issue can be fixed by `phpcbf` tool
- :heavy_check_mark:  yes
- ✖️ no
- :heavy_check_mark: ✖️ partially

| Sniff Description | Fixable |
| --- | --- |
| 1️⃣0️⃣ **ERRORS**  - _Critical issues_ |
|**Generic.Functions.CallTimePassByReference** Ensures that variables are not passed by reference when calling a function. | ✖️ |
|**Generic.PHP.CharacterBeforePHPOpeningTag** Checks that the opening PHP tag is the first content in a file. | ✖️ |
|**Generic.PHP.DeprecatedFunctions** Discourages the use of deprecated PHP functions. | ✖️ |
|**Generic.PHP.NoSilencedErrors** Throws an error or warning when any code prefixed with an asperand is encountered. | ✖️ |
|**Generic.PHP.Syntax** Ensures PHP believes the syntax is clean. | ✖️ |
|**Magento.Legacy.MageEntity** Detects typical Magento 1.x (`Mage::`) classes constructions. | ✖️ |
|**Magento.NamingConvention.ReservedWords** Validates that class name is not reserved word. | ✖️ |
|**Magento.PHP.FinalImplementation** Magento is a highly extensible and customizable platform. The use of `final` classes and methods is prohibited. | ✖️ |
|**Magento.PHP.Goto** Detects use of `goto`. | ✖️ |
[...]

PR #37, opened 2019-02-11
Assign severity to each sniff

lenaorobei:
- changed rules severities based on #14

### DIFF
--- a/Magento/Sniffs/PHP/DiscouragedFunctionSniff.php
+++ b/Magento/Sniffs/PHP/DiscouragedFunctionSniff.php
@@ -235,7 +235,7 @@ class DiscouragedFunctionSniff extends ForbiddenFunctionsSniff
      *
      * @return void
      */
-    protected function addError($phpcsFile, $stackPtr, $function, $pattern = null)
+    protected function addError($phpcsFile, $stackPtr, $function, $pattern=null)
     {
         $data = [$function];
         $warningMessage = 'The use of function %s() is discouraged';
--- a/Magento/Sniffs/PHP/DiscouragedFunctionSniff.php
+++ b/Magento/Sniffs/PHP/DiscouragedFunctionSniff.php
@@ -235,7 +235,7 @@ class DiscouragedFunctionSniff extends ForbiddenFunctionsSniff
      *
      * @return void
      */
-    protected function addError($phpcsFile, $stackPtr, $function, $pattern=null)
+    protected function addError($phpcsFile, $stackPtr, $function, $pattern = null)
     {
         $data = [$function];
         $warningMessage = 'The use of function %s() is discouraged';
--- a/Magento/ruleset.xml
+++ b/Magento/ruleset.xml
@@ -5,158 +5,59 @@
     <!-- File extensions to be checked. -->
     <arg name="extensions" value="php,phtml"/>
 
-    <rule ref="Generic.Arrays.DisallowLongArraySyntax">
-        <severity>8</severity>
-        <type>warning</type>
-    </rule>
-    <rule ref="Generic.Classes.DuplicateClassName">
-        <severity>8</severity>
-    </rule>
-    <rule ref="Generic.CodeAnalysis.ForLoopShouldBeWhileLoop">
-        <severity>6</severity>
-    </rule>
-    <rule ref="Generic.CodeAnalysis.ForLoopWithTestFunctionCall">
-        <severity>6</severity>
-    </rule>
-    <rule ref="Generic.CodeAnalysis.JumbledIncrementer">
-        <severity>8</severity>
-    </rule>
-    <rule ref="Generic.CodeAnalysis.UnconditionalIfStatement">
-        <severity>6</severity>
-    </rule>
-    <rule ref="Generic.CodeAnalysis.UnusedFunctionParameter">
-        <severity>6</severity>
-    </rule>
-    <rule ref="Generic.CodeAnalysis.UselessOverridingMethod">
-        <severity>6</severity>
-    </rule>
+    <!-- Severity 10 errors: Critical code issues. -->
     <rule ref="Generic.Functions.CallTimePassByReference">
         <severity>10</severity>
-    </rule>
-    <rule ref="Generic.Metrics.CyclomaticComplexity">
-        <severity>8</severity>
-        <type>warning</type>
-    </rule>
-    <rule ref="Generic.Metrics.NestingLevel">
-        <severity>8</severity>
+        <type>error</type>
     </rule>
     <rule ref="Generic.PHP.CharacterBeforePHPOpeningTag">
         <severity>10</severity>
+        <type>error</type>
         <exclude-pattern>*.phtml</exclude-pattern>
     </rule>
     <rule ref="Generic.PHP.DeprecatedFunctions">
         <severity>10</severity>
+        <type>error</type>
     </rule>
-    <rule ref="Generic.PHP.NoSilencedErrors.Discouraged">
-        <severity>8</severity>
+    <rule ref="Generic.PHP.NoSilencedErrors">
+        <severity>10</severity>
+        <type>error</type>
     </rule>
     <rule ref="Generic.PHP.Syntax">
         <severity>10</severity>
-    </rule>
-    <rule ref="Magento.Annotation">
-        <severity>6</severity>
-    </rule>
-    <rule ref="Magento.Classes.ObjectInstantiation">
-        <severity>8</severity>
-    </rule>
-    <rule ref="Magento.Classes.ObjectManager">
-        <severity>8</severity>
-    </rule>
-    <rule ref="Magento.CodeAnalysis.EmptyBlock">
-        <severity>8</severity>
-        <type>warning</type>
-    </rule>
-    <rule ref="Magento.Exceptions.DirectThrow">
-        <severity>8</severity>
-    </rule>
-    <rule ref="Magento.Exceptions.Namespace">
-        <severity>8</severity>
-    </rule>
-    <rule ref="Magento.Files.LineLength">
-        <properties>
-            <property name="lineLimit" value="120"/>
-            <property name="absoluteLineLimit" value="120"/>
-        </properties>
-        <type>warning</type>
-        <severity>6</severity>
+        <type>error</type>
     </rule>
     <rule ref="Magento.Legacy.MageEntity">
         <severity>10</severity>
-    </rule>
-    <rule ref="Magento.NamingConvention.InterfaceName">
-        <severity>6</severity>
+        <type>error</type>
     </rule>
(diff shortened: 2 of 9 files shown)


magento-cicd2:
[CLA assistant check](https://cla.dev.magento.com/magento/magento-coding-standard?pullRequest=37) <br/>Thank you for your submission, we really appreciate it. Like many open source projects, we ask that you sign our [Contributor License Agreement](https://cla.dev.magento.com/magento/magento-coding-standard?pullRequest=37) before we can accept your contribution.<br/>

Magento/Sniffs/PHP/DiscouragedFunctionSniff.php
```diff
@@ -235,7 +235,7 @@ class DiscouragedFunctionSniff extends ForbiddenFunctionsSniff
      *
      * @return void
      */
-    protected function addError($phpcsFile, $stackPtr, $function, $pattern = null)
+    protected function addError($phpcsFile, $stackPtr, $function, $pattern=null)
```
orlangur: Where such spaces removal around `=` comes from?

Magento/ruleset.xml
```diff
@@ -5,202 +5,402 @@
...
+        <type>warning</type>
     </rule>
-    <rule ref="Squiz.PHP.GlobalKeyword">
-        <severity>8</severity>
+    <rule ref="PSR2.Files.EndFileNewline">
+        <severity>6</severity>
         <type>warning</type>
     </rule>
-    <rule ref="Squiz.PHP.NonExecutableCode">
+    <rule ref="PSR2.Methods.FunctionCallSignature">
```
orlangur: Revising PSR rules severities is not a good idea at all.

Magento/Sniffs/PHP/DiscouragedFunctionSniff.php
lenaorobei: You're right. I missed that part of PSR2 standard. Will fix.

Magento/ruleset.xml
lenaorobei: Please provide more details why it is not a good idea.

orlangur:
Hi @lenaorobei, not sure if it's related to this PR. I always wanted to have Magento codebase following PSR-2 as strictly as possible. Some warnings, like leading underscores, are really hard to fix all over the codebase.

Do we have a separate ruleset for new files planned? @paliarush mentioned something like this. In such ruleset we would, for example, convert some warnings to errors just to have new code even cleaner in terms of PSR-2.

lenaorobei:
@orlangur 
>I always wanted to have Magento codebase following PSR-2 as strictly as possible.

This is definitely what we want to archive and what we are aiming to. The purpose of this coding standard is validation not only Magento codebase but also developers extensions on Marketplace, so we work iteratively in that regard.

I reviewed ruleset and made it follow PSR2. I understand that severities do not make sense for core, but we need them for Marketplace technical review. We even have [a scoring approach proposal](https://github.com/magento/magento-coding-standard/issues/14#issuecomment-456533044), which should improve the overall code quality of Marketplace extensions.

Magento/Sniffs/PHP/DiscouragedFunctionSniff.php
lenaorobei: I fixed some rules to be compatible with PSR2. 
Anyway, severities are all about Marketplace checks so you don't have to worry about Magento core. For core all findings will lead to build failure.

orlangur:
@lenaorobei,

> severities do not make sense for core, but we need them for Marketplace technical review

It means that it should be a separate standard and there is no need to spoil core coding standard :)

Of course, we shouldn't have two unrelated standards, like it was with MEQP and core standard, but the standard for Marketplace must be based on core one. Some new rules may be introduced in it which are not applicable to core, some severities adjusted to fulfill needs of Marketplace extension review.

But we cannot fulfill contradicting requirements in the scope of a single standard.

Magento/ruleset.xml
```diff
@@ -5,202 +5,438 @@
...
+    <!-- Severity 9 warnings: Possible security issues. -->
+    <rule ref="Magento.Security.IncludeFile">
+        <severity>9</severity>
+        <type>warning</type>
+    </rule>
+    <rule ref="Magento.Security.Superglobal.SuperglobalUsageWarning">
+        <severity>9</severity>
+        <type>warning</type>
+    </rule>
+    <rule ref="Magento.Security.XssTemplate">
```
melnikovi: I propose we increase severity here, we should be very strict about this.

Magento/ruleset.xml
```diff
@@ -5,202 +5,438 @@
...
-    <rule ref="Magento.Templates.ThisInTemplate">
-        <include-pattern>*.phtml</include-pattern>
+
+    <!-- Severity 6 warnings: Code style issues. -->
+    <rule ref="Generic.ControlStructures.InlineControlStructure">
         <severity>6</severity>
+        <type>warning</type>
     </rule>
-    <rule ref="Magento.Translation.ConstantUsage">
+    <rule ref="Generic.Files.ByteOrderMark">
```
melnikovi: Propose to increases severity to 9. We can rename section to "Possible security and issues that may cause bugs".

Magento/ruleset.xml
```diff
@@ -5,202 +5,438 @@
...
-    <rule ref="Magento.Security.XssTemplate">
-        <include-pattern>*.phtml</include-pattern>
-        <severity>8</severity>
+    <rule ref="Squiz.PHP.GlobalKeyword">
+        <severity>7</severity>
+        <type>warning</type>
     </rule>
-    <rule ref="Magento.Strings.ExecutableRegEx">
-        <severity>10</severity>
+    <rule ref="Squiz.PHP.NonExecutableCode">
```
melnikovi: Propose to increase severity to 9.

Magento/ruleset.xml
```diff
@@ -5,202 +5,438 @@
...
+    </rule>
+    <rule ref="Magento.CodeAnalysis.EmptyBlock">
+        <severity>7</severity>
+        <type>warning</type>
+    </rule>
+    <rule ref="Magento.Exceptions.Namespace">
+        <severity>7</severity>
+        <type>warning</type>
+    </rule>
+    <rule ref="Magento.PHP.DiscouragedFunction">
```
melnikovi: Per our discussion, let's increase severity to 9 because it checks for functions that may lead to security issues.

Magento/ruleset.xml
```diff
@@ -5,202 +5,438 @@
...
     </rule>
     <rule ref="Magento.NamingConvention.InterfaceName">
-        <severity>6</severity>
-    </rule>
-    <rule ref="Magento.NamingConvention.ReservedWords">
-        <severity>10</severity>
+        <severity>8</severity>
+        <type>warning</type>
     </rule>
     <rule ref="Magento.PHP.DateTime">
```
melnikovi: Propose increase severity to 9 as it may cause bugs.

Magento/ruleset.xml
```diff
@@ -5,202 +5,438 @@
...
     </rule>
-    <rule ref="Magento.Annotation">
-        <severity>6</severity>
+    <rule ref="Squiz.PHP.Eval">
+        <severity>10</severity>
+        <type>error</type>
+    </rule>
+
+    <!-- Severity 9 warnings: Possible security issues. -->
+    <rule ref="Magento.Security.IncludeFile">
```
melnikovi: Propose increase severity to 9 as it may cause security issues.

Magento/ruleset.xml
lenaorobei: Unfortunately we cannot restrict this rule right now since its _possible_ issue. We will work iteratively in that regard.

Magento/ruleset.xml
lenaorobei: I assume you mean "increase to 10". Done.

lenaorobei:
@melnikovi changed to proposed severities.

Updated [WIKI page](https://github.com/magento/magento-coding-standard/wiki/Magento-Marketplace-Extensions-Verification) as well.

Magento/ruleset.xml
```diff
@@ -5,202 +5,438 @@
...
+        <severity>10</severity>
+        <type>error</type>
+        <exclude-pattern>*.phtml</exclude-pattern>
     </rule>
+    <rule ref="Squiz.PHP.Eval">
+        <severity>10</severity>
+        <type>error</type>
+    </rule>
+
+    <!-- Severity 9 warnings: Possible security and issues that may cause bugs. -->
```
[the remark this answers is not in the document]
lenaorobei: PHPStorm will show all issues regardless of severities. I didn't get what you mean here.
>split the rules with different severity into different rulesets

Severities are for Marketplace technical review only. What is the purpose of splitting into different rulesets here?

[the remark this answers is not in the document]
paliarush: Do we expect to see all issues with any severity highlighted by PhpStorm? Also, this ruleset will be used for Magento core, right?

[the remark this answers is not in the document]
lenaorobei: Yes, we expect to see all issues. 
Yes, this ruleset will be used for Magento core.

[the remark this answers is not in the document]
paliarush: I believe we agreed to move out potentially false-positive rules to separate ruleset to be used with incremental checks.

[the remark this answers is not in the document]
lenaorobei: Right now all checks are used with incremental approach. The same behaviour will preserved for the first release. We can discuss false-positive rulesets for future improvements.

Magento/Sniffs/PHP/DiscouragedFunctionSniff.php
orlangur: @lenaorobei I wouldn't worry about Magento core is there was a separate Marketplace coding standard based on core one. What is the reason to mix such totally different concerns?

lenaorobei: @orlangur why do you think so? Extension developers should follow the same rules as core developers.

Magento/ruleset.xml
```diff
@@ -305,6 +306,7 @@
     <rule ref="PSR2.Classes.PropertyDeclaration">
         <severity>6</severity>
         <type>warning</type>
+        <exclude name="PSR2.Classes.PropertyDeclaration.Underscore"/>
```
paliarush: Why are these excluded? We do not want any variables and method with underscore prefixes.

lenaorobei: It's because a lot of legacy code in Magento. When extension developers extend some classes they need to suppress such findings.

paliarush: We need some strategy for the new code to enable such rules there.

Issue #26, opened 2019-01-31
[New Rule] Getters SHOULD NOT change the state of an object


lenaorobei:
### Rule
Getters SHOULD NOT change the state of an object.

### Reason
Source: [Magento Technical Guidelines](https://devdocs.magento.com/guides/v2.3/coding-standards/technical-guidelines.html).

### Implementation
Check if method name starts with `get` and find class member assignment operation inside this method.
```
<severity>8</severity>
<type>warning</type>
```

larsroettig:
Hi @paliarush and @lenaorobei,
I think we get too many false positives it is very hard to Implementing this with tokens maybe with phpstan it is possible to implement this check.

PR #57, opened 2019-03-08
#26: Add Sniff for Getters not change state

larsroettig:
see #26

### DIFF
--- a/Magento/Sniffs/Functions/GetterStateSniff.php
+++ b/Magento/Sniffs/Functions/GetterStateSniff.php
@@ -74,7 +74,19 @@ class GetterStateSniff implements Sniff
             }
 
             if ($isObjectScope === true && array_key_exists($token['code'], Tokens::$assignmentTokens)) {
-                $phpcsFile->addWarning($this->warningMessage, $i, $this->warningCode);
+
+                $isWrappedByIf = false;
+                // Detect if the property warped by an if tag.
+                $ifTag = $phpcsFile->findPrevious(T_IF, $i);
+                if ($ifTag !== false) {
+                    $open = $tokens[$ifTag]['scope_opener'];
+                    $close = $tokens[$ifTag]['scope_closer'];
+                    $isWrappedByIf = $open <= $i && $close >= $i;
+                }
+
+                if ($isWrappedByIf === false) {
+                    $phpcsFile->addWarning($this->warningMessage, $i, $this->warningCode);
+                }
             }
         }
     }
--- a/Magento/Sniffs/Functions/GetterStateSniff.php
+++ b/Magento/Sniffs/Functions/GetterStateSniff.php
@@ -1,0 +1,81 @@
+<?php
+/**
+ * Copyright © Magento. All rights reserved.
+ * See COPYING.txt for license details.
+ */
+
+namespace Magento\Sniffs\Functions;
+
+use PHP_CodeSniffer\Sniffs\Sniff;
+use PHP_CodeSniffer\Files\File;
+use PHP_CodeSniffer\Util\Tokens;
+
+/**
+ * Detects getter that change the state of an object.
+ */
+class GetterStateSniff implements Sniff
+{
+
+    /**
+     * String representation of warning.
+     *
+     * @var string
+     */
+    protected $warningMessage = 'Getters SHOULD NOT change the state of an object.';
+
+    /**
+     * Warning violation code.
+     *
+     * @var string
+     */
+    protected $warningCode = 'GetterState';
+
+    /**
+     * @inheritDoc
+     */
+    public function register()
+    {
+        return [T_FUNCTION];
+    }
+
+    /**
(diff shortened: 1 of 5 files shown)


lenaorobei:
Could you please commit `ruleset.xml` as well.

larsroettig:
@lenaorobei done 😸

Magento/Sniffs/Functions/GetterStateSniff.php
```diff
@@ -0,0 +1,81 @@
...
+    }
+
+    /**
+     * @inheritDoc
+     */
+    public function process(File $phpcsFile, $stackPtr)
+    {
+        $methodName = $phpcsFile->getDeclarationName($stackPtr);
+
+        if ($methodName === null || strpos($methodName, 'get') === false) {
```
lenaorobei: Isn't it more reliable to check `strpos($methodName, 'get') !==0`?

lenaorobei:
Need to take property caching into account.

larsroettig:
Hi @lenaorobei, 
if we change this check to the only method should change the class property. We break a very typical  *Facade design pattern* this will not work any more or marked as wrong but it can be valid.

### Conditions:
-  My SizeService has very complex logic ``set`` and ``get`` sizes very expensively.

#### Possbily Solutions

1.  We should check for the `if` block is in there assignment this is okay caching reason. 
2. We ignore with `set` or  ` __construct`prefix for property assignments check.

See my Example: 

```php
<?php

namespace Foo\Bar;

class SizeService
{
    /**
     * @return int
     */
    public function getSize()
    {
        // some implementation
        return 12;
    }

    /**
     * @param $size
     * @return void
     */
    public function setSize($size)
    {
        // some implementation
    }
}

interface SizeServiceInterface
{
    /**
     * @return int
     */
    public function getSize();

    /**
     * @param $size
     * @return void
     */
    public function setSize($size);
}


class SizeServiceFacade implements SizeServiceInterface
{
    /**
     * @var int
     */
    private $size;

    /**
     * @var SizeService
     */
    private $service;

    /**
     * @param SizeService $service
     */
    public function __construct(SizeServiceInterface $service)
    {
        $this->service = $service;
    }

    /**
     * @return int
     */
    public function getSize()
    {
        if ($this->size === null) {
            $this->size = $this->service->getSize();
        }

        return $this->size;
    }

    /**
     * @return void
     */
    public function setSize($size)
    {
        $this->size = $size;
        $this->service->setSize($size);
    }
}
```

larsroettig:
Hi @paliarush and @lenaorobei,
I think we get too many false positives it is very hard to Implementing this with tokens maybe with phpstan it is possible to implement this check.


**Testcase:**
```php

<?php

namespace Foo\Bar;


abstract class Bar
{
    public static $foobar = 100;
}

class Foo extends Bar
{
    /**
     * @var int
     */
    private static $staticProperty;

    /**
     * @var int
     */
    private $property;

    /**
     * @return int
     */
    public function getStaticProperty()
    {
        self::$staticProperty = 12;
        static::$staticProperty -= 12;
        self::$staticProperty .= 12;
        return self::$staticProperty;
    }

    /**
     * @return int
     */
    public function getProperty()
    {
        if (true) {

        }

        $this->property = 1223;
        return $this->property;
    }

    /**
     * @return int
     */
    public function getPropertyCached()
    {
        if ($this->property === null) {
            $this->property = 1223;
        }

        return $this->property;
    }

    public function getPropertyLocal()
    {
        $local = $this->property;
        $localArray = [
            'payment' => [
                'test' => [
                    'isActive' => $this->config->isActive(),
                    'title' => $this->config->getTitle()
                ]
            ]
        ];
        return $this->property;
    }

    private function getSalesChannelForOrder($order)
    {
        $websiteId = (int)$order->getStore()->getWebsiteId();
        $websiteCode = $this->websiteRepository->getById($websiteId)->getCode();

        return $this->salesChannelFactory->create([
            'data' => [
                'type' => '',
                'code' => $websiteCode
            ]
        ]);
    }

    const MODE_AUTO = 0;

    const MODE_MANUAL = 1;

    public function getOptionsArray()
    {
        return [
            self::MODE_AUTO => __('Automatically'),
            self::MODE_MANUAL => __('Manually')
        ];
    }

    public function testigetFoo()
    {
        $this->property = 1223;
        return $this->property;
    }

    /**
     * @return int
     */
    public function normalMethod()
    {
        $localVariable = 12;
        return $localVariable;
    }

    public function getStorageModel($storage = null, $params = [])
    {
        if ($storage === null) {
            $storage = $this->_coreFileStorage->getCurrentStorageCode();
        }

        switch ($storage) {
            case self::STORAGE_MEDIA_FILE_SYSTEM:
                $model = $this->_fileFactory->create();
                break;
            default:
                return false;
        }

        if (isset($params['init']) && $params['init']) {
            $model->init();
        }

        return $model;
    }
}

$d = function ($test) {
    $test = 123;
};
```

PR #59, opened 2019-03-11
Proxies and interceptors

### LINKED ISSUE #18: [New Rule] Proxies and interceptors MUST never be explicitly requested in constructors
### Rule
Proxies and interceptors MUST never be explicitly requested in constructors.

Source - [Magento Technical Guidelines](https://devdocs.magento.com/guides/v2.3/coding-standards/technical-guidelines.html)

### Description
Proxies are used for lazy-loading. Details on [DevDocs](https://devdocs.magento.com/guides/v2.3/extension-dev-guide/proxies.html).

Proxies should be configured via `di.xml` file.
```
<type name="FastLoading">
    <arguments>
        <argument name="slowLoading" xsi:type="object">SlowLoading\Proxy</argument>
    </arguments>
</type>
```

Plugins (interceptors) modify the behavior of public class functions by intercepting a function call and running code before, after, or around that function call. Details on [DevDocs](https://devdocs.magento.com/guides/v2.3/extension-dev-guide/plugins.html).

Plugins should be declared via `di.xml` file.

```
<config>
    <type name="{ObservedType}">
      <plugin name="{pluginName}" type="{PluginClassName}" sortOrder="1" disabled="false" />
    </type>
</config>
```
### Implementation
Subscribe to the `T_CLASS` token and analyse `__constructor` arguments.
```xml
<severity>8</severity>
<type>warning</type>
```

maderlock:
Fix #18 New rule to ensure proxies and interceptors are never explicitly requested in constructors


Magento/Sniffs/Classes/ConstructorProxyInterceptorSniff.php
```diff
@@ -0,0 +1,267 @@
...
+class ConstructorProxyInterceptorSniff implements Sniff
+{
+    const CONSTRUCT_METHOD_NAME = '__construct';
+
+    /**
+     * String representation of warning.
+     *
+     * @var string
+     */
+    protected $warningMessage = 'Proxies and interceptors MUST never be explicitly requested in constructors.';
```
paliarush: It would be better to split this sniff into two sniffs to comply with single responsibility principle.

lenaorobei: I have concerns regarding your comment:
1. This sniff covers one statement from Technical Guidelines and thats why it sounds logical to me to have this one rule.
2. Splitting into two sniffs doesn't sound good in performance prospective (make same check twice).
3. On the other hand, making it more generic sounds good to me.  The name can be changed to `DiscouragedDependenciesSniff` (or something like this) and `$warnNames` property can be public and will be replaceable via `ruleset.xml`.

Magento/Sniffs/Classes/ConstructorProxyInterceptorSniff.php
```diff
@@ -0,0 +1,267 @@
...
+     * @var array Aliases of proxies or plugins from use statements
+     */
+    private $aliases = [];
+
+    /**
+     * @var array Terms to search for in variables and namespaces
+     */
+    private $warnNames = [
+        'proxy',
+        'plugin'
```
paliarush: Probably the task was not clear enough, the expectation was to prevent `Interceptor` classes being referenced directly. Interceptors are auto-generated classes. 
It may make sense to run these tests against Magento codebase and see if Plugin reference detection also makes sense.

Magento/Tests/Classes/ConstructorProxyInterceptorUnitTest.2.inc
```diff
@@ -0,0 +1,54 @@
...
+
+use \Vendor\Module\Proxy\AnyName as ProxAlias;
+use \Exception as SafeAlias;
+
+class AnyName {}
+
+/**
+ * // Rule find: constructor use of proxy class
+ */
+class Foo
```
paliarush: Proxy classes are the one which have `\Proxy` suffix. Interceptor classes always have `\Interceptor` suffix. These can be accurately detected and should raise error.


Plugin classes have arbitrary names and thus are difficult to catch. These rules may have accuracy issues and can raise warnings.

maderlock:
To confirm from above:
1. Should make naming more generic and use commenting to make clear what types of class we're warning about including in constructors explicitly
2. Plugins should be removed from this sniff as this was not in the original ticket
3. Interceptors should be added instead.

Magento/Tests/Classes/ConstructorProxyInterceptorUnitTest.2.inc
maderlock: As this is based on names of constructor types and variables there will be a false-positive issue. In discussion with @lenaorobei it was felt that keeping these as warnings was of utility despite this.

Magento/Sniffs/Classes/ConstructorProxyInterceptorSniff.php
maderlock: Makes sense. I'll add interceptors first and then look at plugins as you suggest.

Magento/Sniffs/Classes/ConstructorProxyInterceptorSniff.php
maderlock: > 3. On the other hand, making it more generic sounds good to me. The name can be changed to DiscouragedDependenciesSniff (or something like this) and $warnNames property can be public and will be replaceable via ruleset.xml.

I do like the sound of the latter - I didn't realise that this could be done. Nice :)

lenaorobei:
@maderlock 
> move exact types matched into ruleset.xm

Let them be in the sniff code. Ruleset will be an extension point (if needed in future).

@paliarush do you have additional comments?

lenaorobei:
@maderlock thanks for your contribution!

I just had quick discussion with @paliarush and we agreed following:
1. This sniff should cover `proxy` and `interceptor` cases only. For plugins i'll create separate task.
2. Please change the severity to `10` and type to `error`.
3. Use more generic sniff name.
4. Please add how to fix this issue to the error message: "Proxies and interceptors MUST never be explicitly requested in constructors. Remove redundant `Interceptor`/`Proxy` suffix" (need to dynamically detect this suffix).

maderlock:
> 2. Please change the severity to `10` and type to `error`.

I'll do as you suggest, though I'm still a little concerned about false positives given the matching of variable names.

Magento/Tests/Classes/ConstructorProxyInterceptorUnitTest.2.inc
lenaorobei: Even is it's bad naming it's worth pointing developers that class name with `Interceptor` and `Proxy` suffix is not a good idea.

maderlock:
With the changes above it looks like we should no longer be detecting proxy/interceptor in variable names. The logic should now be:
* If variable type in constructor signature has a suffix of Proxy or Interceptor (or is aliased to a class matching this pattern) then this produces an error

Should we allow classes with Proxy or Interceptor in the non-suffix position, e.g. /Vendor/Module/Proxy/SomethingElse - or should this be a warning or error?

lenaorobei:
>If variable type in constructor signature has a suffix of Proxy or Interceptor (or is aliased to a class matching this pattern) then this produces an error

Yes, that's right.

>Should we allow classes with Proxy or Interceptor in the non-suffix position, e.g. /Vendor/Module/Proxy/SomethingElse - or should this be a warning or error?

Non-suffix position should be fine. We dont need to trigger any warnings in this case.

To make sure everything works as expected, I suggest running this check against Magento codebase:
`vendor/bin/phpcs --standard=Magento --sniffs=Magento.Classes.ConstructorProxyInterceptor /path/to/magento2/app/code`. It will show possible edge cases.

maderlock:
Sorry for the delay - I've been rather ill. I'll jump back on this when I can. Thanks for bearing with me.

lenaorobei:
@maderlock do you need any help with this PR?

maderlock:
Could not get properties passing from ruleset.xml, so I've had to ignore that change request. Otherwise I've made the requested changes and merged into the updated structure under Magento2 (rather than Magento).

I've tested the rule against the Magento codebase and the only errors that come out are valid ones, so I'm happy that it's working as expected (in addition to the tests passing).

Magento2/Sniffs/Classes/DiscouragedDependenciesSniff.php
```diff
@@ -0,0 +1,281 @@
+<?php
+/**
+ * Copyright © Magento. All rights reserved.
+ * See COPYING.txt for license details.
+ */
+
```
lenaorobei:
```suggestion
```

Magento2/Tests/Classes/DiscouragedDependenciesUnitTest.1.inc
```diff
@@ -0,0 +1,55 @@
...
+
+/**
+ * // Rule find: This is fine
+ */
+class Foo4
+{
+    public function __construct(SafeAlias $other) {
+
+    }
+}
```
lenaorobei:
```suggestion
}

```

Magento2/Tests/Classes/DiscouragedDependenciesUnitTest.2.inc
```diff
@@ -0,0 +1,55 @@
...
+
+/**
+ * // Rule find: This is fine
+ */
+class Foo4
+{
+    public function __construct(SafeAlias $other) {
+
+    }
+}
```
lenaorobei:
```suggestion
}

```

Magento2/Tests/Classes/DiscouragedDependenciesUnitTest.php
```diff
@@ -0,0 +1,47 @@
+<?php
+/**
+ * Copyright © Magento. All rights reserved.
+ * See COPYING.txt for license details.
+ */
+
```
lenaorobei:
```suggestion
```

Magento2/ruleset.xml
```diff
@@ -28,6 +28,10 @@
         <severity>10</severity>
         <type>error</type>
     </rule>
+    <rule ref="Magento2.Classes.DiscouragedDependencies">
+        <severity>10</severity>
+        <type>error</type>
```
lenaorobei:
```suggestion
        <type>error</type>
        <exclude-pattern>*/Test/*</exclude-pattern>
```

maderlock:
Minor suggestions addressed and latest develop merged

Issue #62, opened 2019-03-19
[Enhancement] DiscouragedFunction rule improvement


lenaorobei:
### Description
`Magento2.PHP.DiscouragedFunction` needs to be changed in order to cover different use-cases.

### Expected behavior

The first part of the rule should cover following:
- The use of insecure functions triggers **9 severity warning** (applicable to all codebase).
Examples of insecure functions: `exec`, `md5`, `unserialize`, `assert`.

The first part of the rule should cover following:
- The use of discouraged functions triggers **8 severity warning** (not applicable to `*/Test/*` and `*/lib/*`).
Examples of discouraged functions: `sizeof`, `print_r`, `^socket_.*$`, `^stream_.*$`.

PR #63, opened 2019-03-19
[Enhancement] DiscouragedFunction rule improvement

lenaorobei:
- split DiscouragedFunction into two to take insecure functions into account with higher severity;
- added `<exclude-pattern>*/Test/*</exclude-pattern>` for rules, not applicable to tests and test files;
- resolved magento/magento-coding-standard#62.


Magento2/Sniffs/Functions/DiscouragedFunctionSniff.php
```diff
@@ -20,13 +19,19 @@ class DiscouragedFunctionSniff extends ForbiddenFunctionsSniff
      */
     protected $patternMatch = true;
 
+    /**
+     * If true, an error will be thrown; otherwise a warning.
+     *
+     * @var boolean
+     */
+    public $error = false;
```
paliarush: Please validate the list against https://stackoverflow.com/questions/3115559/exploitable-php-functions and ask security team to review.

Magento2/Sniffs/Security/InsecureFunctionSniff.php
```diff
@@ -0,0 +1,37 @@
...
+ * Detects the use of insecure functions.
+ */
+class InsecureFunctionSniff extends ForbiddenFunctionsSniff
+{
+    /**
+     * If true, an error will be thrown; otherwise a warning.
+     *
+     * @var boolean
+     */
+    public $error = false;
```
paliarush: Why is it a warning?

lenaorobei: It's under `Severity 9 warnings: Possible security and issues that may cause bugs.` section.
We cannot reject extensions because of such functions right now. We will work iteratively to make rules stricter.

Magento2/Sniffs/Functions/DiscouragedFunctionSniff.php
lenaorobei: Added.

Issue #20, opened 2019-01-29
[New Rule] Slow array SHOULD NOT be used in loop


lenaorobei:
### Rule

The use of slow array functions (`array_merge`) in loop is discouraged.

### Reason
Merging arrays in a loop is slow and causes high CPU usage. [Some benchmarking](https://github.com/kalessil/phpinspectionsea/issues/138#issuecomment-279457133).

**Bad example:**
``` php
    $options = [];
    foreach ($configurationSources as $source) {
        // code here
        $options = array_merge($options, $source->getOptions());
    }
```

**Good example:**
``` php
    $options = [[]];
    foreach ($configurationSources as $source) {
        // code here
        $options[] = $source->getOptions();
    }

    // PHP 5.6+
    $options = array_merge(...$options);
```

### Implementation
- Subscribe to `T_STRING` token and check if it's content is `array_merge`. 
- Try to find outer loop.
- If loop is found rase a warning (need to discuss additional logic).

```
<severity>7</severity>
<type>warning</type>
```

lenaorobei:
We can throw a warning with incremental approach.

Vinai:
Mixed feelings. I often use merge in a reduce function to avoid changing local state (and save a line of code).
Hope PHP optimizes for more functional patterns in future, now that FP is becoming mainstream.
Even Java now has optimized map/reduce/filter chains, and more and more languages support STM.
If PHP follows, this sniff would be outdated.
Maybe sniffs should focus on algorithmic complexity instead of Runtime implementation details.

shochdoerfer:
Slow is a matter of definition :) In a loop with just a few iterations, you most likely won't notice that much of a difference.

Vinai:
Exactly. That’s why focusing on big O algorithmic complexity seems like a good idea since that’s assumes scalability in the worst case scenario.
Stating this I feel stupid because I’m sure you are well aware.

larsroettig:
Benmarks:

- https://gist.github.com/Ocramius/8399625#file-output-txt

- https://github.com/kalessil/phpinspectionsea/blob/master/docs/performance.md#slow-array-function-used-in-loop

- https://github.com/kalessil/phpinspectionsea/issues/138#issuecomment-279457133

lenaorobei:
@buskamuza @kandy 

Need you input here.

See related PR magento/magento-coding-standard#72

kandy:
For me, it makes sense to have this rule

larsroettig:
[...]

PR #72, opened 2019-03-21
#20  Impelement ArrayMerge sniff in foreach

larsroettig:
Implementation for #20 

### Currently we have many findings in 2.3 develop

```

bin/phpcs --standard=Magento2 --sniffs=Magento2.Performance.ForeachArrayMerge  app/code app/lib
```

###  Magento 2 findings with current Implementation 
```
------------------------------------------------------------------------------------------------------
A TOTAL OF 0 ERRORS AND 280 WARNINGS WERE FOUND IN 212 FILES
------------------------------------------------------------------------------------------------------
```


lenaorobei:
I checked some findings and they look false-positive to me. Probably, they require serious code refactoring, not just moving `array_merge` outside the loop.
Need more thoughts here to understand if we really want to warn developers about any case of usage  `array_merge` inside the loop.

Magento2/Sniffs/Performance/ForeachArrayMergeSniff.php
```diff
@@ -0,0 +1,72 @@
+<?php
+/**
+ * Copyright © Magento. All rights reserved.
+ * See COPYING.txt for license details.
+ */
+
```
lenaorobei:
```suggestion
```

lenaorobei:
```suggestion
```

Magento2/Tests/Performance/ForeachArrayMergeUnitTest.inc
```diff
@@ -0,0 +1,43 @@
...
+    }
+}
+
+$selectBuilder = new SelectBuilder();
+
+foreach ([] as $collection) {
+    foreach ($configurationSources as $source) {
+        $selectBuilder->setColumns(array_merge($selectBuilder->getColumns(), $source));
+    }
+}
```
lenaorobei:
```suggestion
}

```

buskamuza:
We should accept it, but include explanation of how to fix the problematic code. Think about a catalog of recommendations (either here in wiki or in devdocs).

orlangur:
@larsroettig @buskamuza 

> A TOTAL OF 0 ERRORS AND 280 WARNINGS WERE FOUND IN 212 FILES

So how this rule can be useful?

Issue #21, opened 2019-01-29
[New Rule] No try block detected when processing system resources


lenaorobei:
### Rule
If a method uses system resources (such as files, sockets, streams, etc), the code MUST be wrapped with a try block and the corresponding finally block. In the finally sections, all resources SHOULD be properly released.

### Reason
Source: [Magento Technical Guidelines](https://devdocs.magento.com/guides/v2.3/coding-standards/technical-guidelines.html).

### Implementation
- Subscribe to `T_STRING` token.
- Prepare the list of build in PHP functions that use resources or alternatively functions prefixes (`stream_`, `socket_`, etc)
- Check if token content is `in_array` of  system functions.
- Raise a warning if no outer `try` statement detected.

lenaorobei:
@magento I am working on it

PR #77, opened 2019-03-31
#21: Impelement rule

larsroettig:
see #21


Magento2/Tests/Exceptions/TryProcessSystemResourcesUnitTest.inc
```diff
@@ -0,0 +1,45 @@
...
+    public function handleExpection()
+    {
+        try {
+            $strChar = stream_get_contents(STDIN, 1);
+        } catch (Exception $exception) {
+
+        }
+    }
+
+    public function exetcute()
```
joshuaadickerson: Typo? `execute`

Magento2/Sniffs/Exceptions/TryProcessSystemResourcesSniff.php
```diff
@@ -0,0 +1,93 @@
+<?php
+/**
+ * Copyright © Magento. All rights reserved.
+ * See COPYING.txt for license details.
+ */
+
```
lenaorobei:
``` suggestion
```

output.txt
```diff
@@ -0,0 +1,4 @@
+ERROR: The specified sniff code "Magento2.Exceptions" is invalid
```
lenaorobei: Looks like this file us redundant.

Magento2/Sniffs/Exceptions/TryProcessSystemResourcesSniff.php
```diff
@@ -0,0 +1,93 @@
...
+ * Detects no try block detected when processing system resources
+ */
+class TryProcessSystemResourcesSniff implements Sniff
+{
+    /**
+     * String representation of warning.
+     *
+     * @var string
+     */
+    protected $warningMessage = 'All resources SHOULD be properly released.';
```
lenaorobei: Please change message to: `The code MUST be wrapped with a 'try' block if the method uses system resources `.

Magento2/Tests/Exceptions/TryProcessSystemResourcesUnitTest.inc
larsroettig: ups 🤦‍♂️

lenaorobei:
Have concerns regarding [Technical Guidelines](https://devdocs.magento.com/guides/v2.3/coding-standards/technical-guidelines.html) that says:
>5.16. If a method uses system resources (such as files, sockets, streams, etc.), the code MUST be wrapped with a try block and the corresponding finally block. In the finally sections, all resources SHOULD be properly released. 

1. Native PHP functions that work with resources (`stream_*`, `socket_*`) do not throw Exceptions or Errors, they mostly return `false` in case of failure and can produce `E_WARNING`. https://www.php.net/manual/en/sockets.errors.php 
https://www.php.net/manual/en/stream.errors.php
2. Magento app will transform `E_WARNING` to the Exception in `Magento\Framework\App\ErrorHandler`. 
3. Magento Framework wrappers that work with resources solve `E_WARNING` by adding `@`, and do not have `try catch` blocks. Examples: 
https://github.com/magento/magento2/blob/2.3-develop/lib/internal/Magento/Framework/Filesystem/Driver/Http.php#L199
https://github.com/magento/magento2/blob/2.3-develop/lib/internal/Magento/Framework/Backup/Filesystem/Rollback/Ftp.php#L93

My question is what is the right way to deal with such cases?
@akaplya @buskamuza @paliarush  @melnikovi @kandy need your input here.

lenaorobei:
@larsroettig this rule was approved on architects discussion https://github.com/magento/architecture/issues/140

I'm going to test it and will provide you feedback.

lenaorobei:
@larsroettig could you please remove `curl_ ` from this sniff to make sure it's in line with Magento Technical Guidelines.

This is an adding new rule, so it will be the part of next major release.
I need to suspend the processing of current PR since the versioning strategy is not clear yet https://github.com/magento/architecture/pull/136

Sorry about that and thanks for understanding.

PR #101, opened 2019-05-22
provide possible solution feedback for discouraged functions

philippsander:


magento-cicd2:
[CLA assistant check](https://cla.dev.magento.com/magento/magento-coding-standard?pullRequest=101) <br/>Thank you for your submission, we really appreciate it. Like many open source projects, we ask that you sign our [Contributor License Agreement](https://cla.dev.magento.com/magento/magento-coding-standard?pullRequest=101) before we can accept your contribution.<br/>

Magento2/Sniffs/Functions/DiscouragedFunctionSniff.php
```diff
@@ -53,7 +53,7 @@ class DiscouragedFunctionSniff extends ForbiddenFunctionsSniff
         '^dcngettext$' => null,
         '^dgettext$' => null,
         '^dio_.*$' => null,
-        '^dirname$' => null,
+        '^dirname$' => '\Magento\Framework\Filesystem\Io\File::getPathInfo [\'dirname\']',
```
paliarush: How about the following:

```suggestion
        '^dirname$' => '\Magento\Framework\Filesystem\Io\File::getPathInfo($filePath)[\'dirname\']',
```

The same relates to `basename`.

philippsander:
I'll fix the CI fails soon

fooman:
I think `dirname` and `basename` is a not a good case for making people use `\Magento\Framework\Filesystem\Io\File::getPathInfo`. The given better uses rely on an implementation detail (the returned array happens to conform to the pathinfo output) with the array having dirname and basename as keys. But this is not really part of any `@api` guarantees.  

I believe it would be better to fix the framework first, either by adding extra methods ie getDirname and getBasename or by adding an optional input to pathinfo($args = null) and then advise people to use those stable implementations.

lenaorobei:
I agree with @fooman here. Neither `\Magento\Framework\Filesystem\Io\File` nor `\Magento\Framework\HTTP\Adapter\Curl` are marked with `@api`, so I'd avoid such recommendation for now.

lenaorobei:
Have to close this PR since we cannot recommend using non-api classes.

PR #104, opened 2019-05-29
REPO-209: [EQP] Increase the severity level to 10 for insecure functions

lenaorobei:
```php
public $forbiddenFunctions = [
        'assert' => null,
        'create_function' => null,
        'exec' => null,
        'md5' => 'improved hash functions (SHA-256, SHA-512 etc.)',
        'passthru' => null,
        'pcntl_exec' => null,
        'popen' => null,
        'proc_open' => null,
        'serialize' => '\Magento\Framework\Serialize\SerializerInterface::serialize',
        'shell_exec' => null,
        'system' => null,
        'unserialize' => '\Magento\Framework\Serialize\SerializerInterface::unserialize',
        'srand' => null,
        'mt_srand' => null,
        'mt_rand' => 'random_int',
    ];
```
Those functions are forbidden in Magento and its use will cause extension rejection on Marketplace.

### DIFF
--- a/Magento2/Sniffs/Security/InsecureFunctionSniff.php
+++ b/Magento2/Sniffs/Security/InsecureFunctionSniff.php
@@ -13,13 +13,6 @@ use PHP_CodeSniffer\Standards\Generic\Sniffs\PHP\ForbiddenFunctionsSniff;
 class InsecureFunctionSniff extends ForbiddenFunctionsSniff
 {
     /**
-     * If true, an error will be thrown; otherwise a warning.
-     *
-     * @var boolean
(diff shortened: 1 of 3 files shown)


gixid192:
I had a case where I need to change the behavior of the export function. The solution was to rewrite the Core class.
Here in the code, the Core team used md5 to generate a unique name: https://github.com/magento/magento2/blob/2.3-develop/app/code/Magento/Ui/Model/Export/ConvertToCsv.php#L68

I understand that there is other ways to generate names, however, is it ok in this case? Just to generate a random name

Issue #105, opened 2019-05-31
[New Rule] Implement sniff for classes and interfaces PHPDoc formatting


lenaorobei:
### Rule
PHPDoc formatting for classes and interfaces.

**Acceptance Criteria:**

- Sniff covers described on [DevDocs requirements](https://devdocs.magento.com/guides/v2.3/coding-standards/docblock-standard-general.html):
> Classes and interfaces should have a short description that is a human-understandable intention of the class.
If the short description adds no additional information beyond what the type name already supplies, the short description must be omitted.
- Rule covered by unit test.

```
<severity>5</severity>
<type>warning</type>
```

lenaorobei:
Hi @udovicic! How is your progress with this ticket? Do you need any help?

udovicic:
Hi @lenaorobei , help no, but more hours in one day definitely. I think I will be done be the end of the week.

Issue #58, opened 2019-03-08
[New Rule] Implement sniff for class properties PHPDoc formatting


lenaorobei:
### Rule
PHPDoc formatting for class properties.

**Acceptance Criteria:**

- Sniff covers described on [DevDocs requirements](https://devdocs.magento.com/guides/v2.3/coding-standards/docblock-standard-general.html):
> Class attributes must have type declaration using `@var` tag.

- Rule covered by unit test.

```
<severity>5</severity>
<type>warning</type>
```

lenaorobei:
Hi @ryantfowler! How is your progress with this ticket? Do you need any help?

ryantfowler:
Hi @lenaorobei , I should have an update by this weekend. I've been tied up with work. Thanks

lenaorobei:
@ryantfowler sorry, unassigning you from this issue since no updates.
Feel free to assign yourself again if you want to continue working on this task. Thanks.

konarshankar07:
Hello @lenaorobei ,
Can we just add `Squiz.Commenting.VariableCommentSniff` in our ruleset?

lenaorobei:
@konarshankar07 need to check if it complies with DevDocs. I checked long time ago, and there was differences. But anyway, it can be taken as a good basis.

konarshankar07:
@lenaorobei  no problem, I'll create new rule for class properties

Issue #54, opened 2019-03-04
[New rule] Implement sniffs for missing rule: @inheritdoc formatting


lenaorobei:
### Rule
As per description in the magento/magento-coding-standard#39
The rule for `@inheritdoc` formatting needs to be implementesd. Docs [here](http://docs.phpdoc.org/guides/inheritance.html).

**Acceptance Criteria:**
- `@inheritdoc` is allowed.
- If the `@inheritdoc` not inline it shouldn't have braces.

larsroettig:
```php
<?php

namespace Foo\Bar;

interface TestInterface
{
    /**
     * Test Function with nice Doc
     *
     * @param $int
     * @param $string
     * @return string
     */
    public function test($int, $string);
}


class Test implements TestInterface
{
    /**
     * @inheritDoc
     */
    public function test($int, $string)
    {
        return '';
    }

    /**
     * Test Function with nice Doc.
     *
     * @param $string
     * @return string
     */
    public function test2($string)
    {
        return '';
    }
}

class  TestExtend extends Test implements TestInterface
{
    /**
     * Test Function with a better Function.
     *
     * {@inheritDoc}
     */
    public function test($int, $string)
    {
        return '';
    }

    /**
     * Sniff will find doc error there because you define inherit wrong.
     * @inheritDoc
     */
    public function test2($string)
    {
        return '';
    }
}
```

lenaorobei:
@larsroettig, few comments to your example:

> The @inheritdoc tag SHOULD NOT be used. 

> The declaration of the return type using the @return tag must only be added if the method return type signature does not supply all necessary information (see below for more information on return types).

lenaorobei:
**Test should NOT fail.**
This example contains negative scenarios:
``` php
<?php

class EverythingIsGoodHere
{
    /**
     * @param string $arg1
     * @param bool $arg2
     */
    private function missingArgumentTypeSoParamNeedsToBeAdded($arg1, $arg2): bool
    {

    }

    /**
     * @return bool
     */
    private function presentArgumentTypeSoAnnotationNotNeededButMissingReturnType(string $arg1, bool $arg2)
    {

    }

    private function presentArgumentAndReturnTypes(string $arg1, bool $arg2): bool
    {

    }

    /**
     * @throws Exception
     */
    private function throwsAnException(): void
    {
        throw new \Exception();
    }

    /**
     * We have an additional wonderful message here.
     *
     * @return void
     */
[...]

PR #106, opened 2019-05-31
Removed Squiz.Commenting.ClassComment since it does not cover all Magento cases

lenaorobei:
As per upcoming updates https://github.com/magento/devdocs/pull/4425 `Squiz.Commenting.ClassComment` does not cover Magento 2 requirements and produces false positive findings.

Related tasks:
- DocBlock for function and methods https://github.com/magento/magento-coding-standard/issues/54
- DocBlock for class properties https://github.com/magento/magento-coding-standard/issues/58
- DocBlock for classes and interfaces https://github.com/magento/magento-coding-standard/issues/105
- DocBlock for constants https://github.com/magento/magento-coding-standard/issues/107

### DIFF
--- a/Magento2/ruleset.xml
+++ b/Magento2/ruleset.xml
@@ -493,13 +493,6 @@
         <severity>5</severity>
         <type>warning</type>
     </rule>
-    <rule ref="Squiz.Commenting.ClassComment">
-        <severity>5</severity>
-        <type>warning</type>
-    </rule>
-    <rule ref="Squiz.Commenting.ClassComment.TagNotAllowed">
-        <severity>0</severity>
-    </rule>
     <rule ref="Squiz.PHP.CommentedOutCode">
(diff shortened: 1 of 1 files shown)


Magento2/ruleset.xml
```diff
@@ -493,13 +493,6 @@
         <severity>5</severity>
         <type>warning</type>
     </rule>
-    <rule ref="Squiz.Commenting.ClassComment">
```
paliarush: Do we have our custom sniff enabled?

lenaorobei: Yes, those Annotation sniffs are still under `dev/tests/static`.
Once new will be implemented here, I'll remove the old ones.
https://github.com/magento/magento2/blob/2.3-develop/dev/tests/static/framework/Magento/ruleset.xml#L12

PR #124, opened 2019-08-03
Implement rule from #105

udovicic:
Implementation of rule from #105 

Additionally, to avoid duplicated code, I have refactored `ConstantsPHPDocFormattingSniff` to make it reusable.


lenaorobei:
@udovicic thanks for opening this PR! It's going to be an awesome improvement, however I have few comments to make it even better 🙂 

1. Let us be more strict with classes and interfaces. What I mean is that `Class should have short description AND this description should be different form the class name`. Both following examples should raise a warning.
``` php
class NotAnErrorHandler
{
}

/**
 * Faulty Handler
 */
class FaultyHandler
{
}
```

2. Please add case with empty DocBlock.
``` php
/**
 *
 */
class FaultyHandler
{
}
```

3. As [per documentation](https://devdocs.magento.com/guides/v2.3/coding-standards/docblock-standard-general.html#documentation-space) some tags are forbidden. It would be great to include these in this check.
`@author ,@category, @package, and @subpackage MUST NOT be used. Documentation is organized with the use of namespaces.`

4. Want to share with you [the core rule](https://github.com/magento/magento2/blob/2.3-develop/dev/tests/static/framework/Magento/Sniffs/Annotation/ClassAnnotationStructureSniff.php) we are trying to replace (because of some bugs and wrong behaviour), it might help with implementation details (like expected messages) and PHPCS out-of-the-box [ClassCommentSniff](https://github.com/squizlabs/PHP_CodeSniffer/blob/master/src/Standards/Squiz/Sniffs/Commenting/ClassCommentSniff.php) rule. This one might help with proper formatting and forbidden tags.

Magento2/Sniffs/Commenting/ClassAndInterfacePHPDocFormattingSniff.php
```diff
@@ -0,0 +1,66 @@
...
+ */
+namespace Magento2\Sniffs\Commenting;
+
+use PHP_CodeSniffer\Sniffs\Sniff;
+use PHP_CodeSniffer\Files\File;
+
+/**
+ * Detects PHPDoc formatting for classes and interfaces.
+ */
+class ClassAndInterfacePHPDocFormattingSniff extends ConstantsPHPDocFormattingSniff implements Sniff
```
lenaorobei: `ClassAndInterfacePHPDocFormattingSniff` should not extend `ConstantsPHPDocFormattingSniff `. Let's move `providesMeaning` to the separate class (aka helper) and call this method from there (it can be even `static`). In future some common methods related to Doc Blocks formatting can be places there.

udovicic:
Hi @lenaorobei ,

I am fine with the rest, but regarding the point 1: Definition says that class **should** have a short description, but that is not a **must**. Furthermore, in [devdocs ](https://devdocs.magento.com/guides/v2.3/coding-standards/docblock-standard-general.html#classes-interfaces) and the[ original issue](https://github.com/magento/magento-coding-standard/issues/105), it actually says:

> If the short description adds no additional information beyond what the type name already supplies, the short description must be omitted.

Therefore, I think that parth should remain as is

lenaorobei:
@udovicic good catch! 🙂 Let me confirm this question with architecture team. My only one concern is that if we will allow no description, nobody will add any description at all.

@paliarush @melnikovi need your input here since you were working on requirements for annotation rules.

Magento2/Sniffs/Commenting/ClassAndInterfacePHPDocFormattingSniff.php
```diff
@@ -0,0 +1,119 @@
...
+    /**
+     * @var PHPDocFormattingValidator
+     */
+    private $PHPDocFormattingValidator;
+
+    /**
+     * @var string[] List of tags that can not be used in comments
+     */
+    private $forbiddenTags = [
+        '@author',
```
ihor-sviziev: Why `@author` is forbidden? As these tests designed to be used not only by Magento - it have to be allowed

lenaorobei: Agree, let's make `$forbiddenTags` public and remove `@author` from the default array. For public properties we can pass values through `ruleset.xml`. As an option, for the core ruleset we will be able to overwrite those values. More details: https://github.com/squizlabs/PHP_CodeSniffer/wiki/Customisable-Sniff-Properties

lenaorobei:
@udovicic I raised my concerns regarding missing PHPDoc Blocks https://github.com/magento/architecture/issues/222 but got confirmation that your approach is correct and this is expected behaviour. Sorry for the confusion.

One more thing. I run your branch against Magento2 `app/code` and found some false-positives.
```
/**
 * @api
 * @since 100.0.2
 */
class Main extends \Magento\Backend\Block\Template
{}
```
This code throws a warning, but it shouldn't.

Magento2/Sniffs/Commenting/ClassAndInterfacePHPDocFormattingSniff.php
udovicic: Done as agreed. Override through `rulset.xml` works as expected

udovicic:
@lenaorobei Not sure how I missed that one, but it has been added as a test case and code has been updated to reflect that.

Everything else should be in there now. I understand that removing description from those classes is not optimal solution, but may come as future refactoring or something, I didn't consider it to be part of the original issue I am trying to resolve here.

lenaorobei:
Hm, that's odd, but now these examples raise warnings.
``` php
/**
 * @api
 */
interface WysiwygModifierInterface
{}
```
``` php
/**
 * @SuppressWarnings(PHPMD.CouplingBetweenObjects)
 */
class GroupRepositoryTest extends \PHPUnit\Framework\TestCase
{}
```
It might be related to the string token after the tag.

Magento2/Sniffs/Commenting/ClassAndInterfacePHPDocFormattingSniff.php
```diff
@@ -0,0 +1,118 @@
...
+            true
+        );
+
+        if ($tokens[$commentStartPtr]['code'] !== T_DOC_COMMENT_OPEN_TAG) {
+            return;
+        }
+
+        if ($this->PHPDocFormattingValidator->providesMeaning($namePtr, $commentStartPtr, $tokens) !== true) {
+            $phpcsFile->addWarning(
+                'If exists, description should add additional information beyond what the type name already supplies.',
```
lenaorobei:
```suggestion
                sprintf(
                    '%s description should contain additional information beyond the name already supplies.',
                    ucfirst($tokens[$stackPtr]['content'])
                ),
```

Magento2/Tests/Commenting/ClassAndInterfacePHPDocFormattingUnitTest.1.inc
```diff
@@ -0,0 +1,79 @@
...
+}
+
+/**
+ * @api
+ * @since 100.0.2
+ */
+class ApiHandler
+{
+
+}
```
lenaorobei:
```suggestion
}

```

Magento2/Tests/Commenting/ClassAndInterfacePHPDocFormattingUnitTest.2.inc
```diff
@@ -0,0 +1,79 @@
...
+}
+
+/**
+ * @api
+ * @since 100.0.2
+ */
+interface ApiHandler
+{
+
+}
```
lenaorobei:
```suggestion
}

```

udovicic:
@lenaorobei I have added the suggested changes and fixed the latest issue. Tag detection was all wrong previously, it should be fine now (I have also added the test cases for it).

PR #129, opened 2019-08-12
#54: PHPDoc formatting

larsroettig:
Ticket: #54


lenaorobei:
@larsroettig fyi, you can reuse https://github.com/magento/magento-coding-standard/blob/develop/Magento2/Sniffs/Commenting/PHPDocFormattingValidator.php

For consistency please change the sniff name to `FunctionsPHPDocFormatting`.

Issue #118, opened 2019-07-28
Space before colon required in phtml files when using alternative syntax for control structures?


hostep:
### Description
Hi

This is more a question or a start of a discusion around the fact that currently a space is required before a colon in phtml files when using the [alternative syntax for control structures](https://www.php.net/manual/en/control-structures.alternative-syntax.php).

So currently Magento code standards expects us to use this format in phtml files:
```php
<?php if ($something) : ?>
    <!-- do something -->
<?php else : ?>
    <!-- do something else -->
<?php endif; ?>
```

I'm wondering why this is required at this moment? Was there a reason for requiring it like this? Or did it just happen because it is the default setting of the PHP_CodeSniffer project?
Because it looks very ugly (personal opinion), other php projects don't enforce this syntax, a lot of php tutorials/samples don't use it, and it's also not part of any php coding standard as far as I'm aware?

I would rather see this format:
```php
<?php if ($something): ?>
    <!-- do something -->
<?php else: ?>
    <!-- do something else -->
<?php endif; ?>
```

This could be changed with this option:
https://github.com/squizlabs/PHP_CodeSniffer/wiki/Customisable-Sniff-Properties#squizcontrolstructurescontrolsignature

I'm seeing a lot of files [being changed recently](https://github.com/magento/magento2/commit/8698a6959f5163ee26b4a900e8749e24ca33e319) in the `2.3-develop` branch whereas in Magento 2.3.2 this syntax wasn't being enforced yet. So it feels like somebody made this decision to enforce this but I can't find the reasoning behind it.

Not sure if this would make a chance to get changed?
No big deal if not, just wanted to bring this up 🙂

Thanks!

chris-pook:
I agree with this, the additional space adds nothing really and is very inconsistent across the codebase.

hostep:
@lenaorobei: any updates here?

If this gets approved, it best happens before Magento 2.3.3 gets released, otherwise a lot of files will get changed to the other format and then again in 2.3.4 to this new proposal (if approved). It would be nice if all changes could happen in a single Magento release if possible.
This makes figuring out what changed in between releases easier as fewer files will have been updated.

Thanks!

lenaorobei:
@hostep I'm fine with this change, but unfortunately we can be done for 2.3.4 and later. We have code freeze for 2.3.3 now.

@paliarush are you ok with it?

PR #132, opened 2019-08-18
No longer require a space before a colon for control structures using…

hostep:
… the alternative syntax

This fixes https://github.com/magento/magento-coding-standard/issues/118

Documentation: https://github.com/squizlabs/PHP_CodeSniffer/wiki/Customisable-Sniff-Properties#squizcontrolstructurescontrolsignature

### DIFF
--- a/Magento2/ruleset.xml
+++ b/Magento2/ruleset.xml
@@ -431,6 +431,9 @@
     <rule ref="Squiz.ControlStructures.ControlSignature">
         <severity>6</severity>
         <type>warning</type>
+        <properties>
+            <property name="requiredSpacesBeforeColon" value="0"/>
+        </properties>
     </rule>
     <rule ref="Squiz.ControlStructures.ControlSignature.NewlineAfterOpenBrace">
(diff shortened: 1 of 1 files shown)


magento-cicd2:
[CLA assistant check](https://cla.dev.magento.com/magento/magento-coding-standard?pullRequest=132) <br/>Thank you for your submission, we really appreciate it. Like many open source projects, we ask that you sign our [Contributor License Agreement](https://cla.dev.magento.com/magento/magento-coding-standard?pullRequest=132) before we can accept your contribution.<br/>

hostep:
I've added a unit test, but for some reason it's not working properly. If somebody could give me a hint why not, that would be appreciated!

It might be because there is no corresponding Sniff added, but does that mean that no unit test is required in this case?

hostep:
I've thought about it some more and it doesn't make much sense to have unit tests for this in the magento-coding-standard repo. The unit tests should be (and are already) part of the PHP_CodeSniffer repo.
So they are removed again over here.

lenaorobei:
@hostep thank you for this improvement.

PR #134, opened 2019-08-22
Implement rule from #131

udovicic:
Implementation of rule from #131

### DIFF
--- a/Magento2/Sniffs/Commenting/ClassAndInterfacePHPDocFormattingSniff.php
+++ b/Magento2/Sniffs/Commenting/ClassAndInterfacePHPDocFormattingSniff.php
@@ -87,6 +87,15 @@ class ClassAndInterfacePHPDocFormattingSniff implements Sniff
             );
         }
 
+        if ($this->PHPDocFormattingValidator->hasDeprecatedWellFormatted($commentStartPtr, $tokens) !== true) {
+            $phpcsFile->addWarning(
+                'Motivation behind the added @deprecated tag MUST be explained. '
+                    . '@see tag MUST be used with reference to new implementation.',
+                $stackPtr,
+                'InvalidDeprecatedTagUsage'
+            );
+        }
+
         $this->validateTags($phpcsFile, $commentStartPtr, $tokens);
     }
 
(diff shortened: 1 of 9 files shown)


lenaorobei:
I run your branch against Magento codebase. Here are some suspicious findings:
```
FILE: app/code/Magento/Ui/Component/Wrapper/Block.php
-------------------------------------------------------------------------------------------------------------------------------------------
FOUND 0 ERRORS AND 2 WARNINGS AFFECTING 2 LINES
-------------------------------------------------------------------------------------------------------------------------------------------
 16 | WARNING | Motivation behind the added @deprecated tag MUST be explained. @see tag MUST be used with reference to new implementation.
 18 | WARNING | Motivation behind the added @deprecated tag MUST be explained. @see tag MUST be used with reference to new implementation.
-------------------------------------------------------------------------------------------------------------------------------------------

FILE: app/code/Magento/Authorizenet/Model/Directpost.php
-------------------------------------------------------------------------------------------------------------------------------------------
FOUND 0 ERRORS AND 2 WARNINGS AFFECTING 2 LINES
-------------------------------------------------------------------------------------------------------------------------------------------
 21 | WARNING | Motivation behind the added @deprecated tag MUST be explained. @see tag MUST be used with reference to new implementation.
 23 | WARNING | Motivation behind the added @deprecated tag MUST be explained. @see tag MUST be used with reference to new implementation.
-------------------------------------------------------------------------------------------------------------------------------------------
```

udovicic:
Hi @lenaorobei,

Thanks for pointing that out, it appears that part of code responsible for detecting DocBlock of constants has a bug, and instead detecting that there is no const DocBlock, it detects one belonging to a class. In both outlined examples, there is a problem with DocBlock as pointed out, but only for classes. Which means that this change is actually ok, problem is one of previous commits. But since is mine either way, I will fix it and report back.

udovicic:
Opens up a room for being lazy and not specifying the alternative. On the other hand, it can't stay the way it is. So I agree with your proposal. Do you want me to change the source or do you want to consult with someone first?

udovicic:
Hi @lenaorobei , just to make sure I fully understand last part, version is not counted as description? So examples that you provided:

`@deprecated 2.3.1` ❌
`@deprecated 2.3.1 Authorize.net is removing all support for this payment method` ✅

Still stand correct?

udovicic:
@lenaorobei I have modified code to reflect latest request. I am not particularly happy with how it looks, but I have no idea how to refactor it to look better either.

PR #139, opened 2019-09-02
Remove ineffective rule

fooman:
I believe this rule is not achieving its intended result. If I understand the intention of the rule md5 should be avoided where collisions can be generated to cause a lapse in security. However in those cases the suggested approach to replace md5 with sha256 I would still consider bad advice. If you are implementing security one should not roll your own.

To judge the effectiveness I ran the following on the code base:
```
git checkout 2.2.0
grep -r --exclude-dir='Test' 'md5(' app/code lib/internal | wc -l
      52

git checkout 2.3-develop
grep -r --exclude-dir='Test' 'md5(' app/code lib/internal | wc -l
      53

grep -r --exclude-dir='Test' 'phpcs:ignore Magento2.Security.InsecureFunction' app/code lib/internal | wc -l
    4
```

In summary we have added more exceptions to the rule than made changes after the rule was introduced.

I do not believe that removing md5 is warranted in a lot of cases (the exceptions allowed confirm this) and that just removing md5 improves security by itself (in other words `hash('string', 'sha256')` is only marginally better than `md5('string')` where really sensitive data is involved, as no one should be handling this directly in the first place.


magento-cicd2:
[CLA assistant check](https://cla.dev.magento.com/magento/magento-coding-standard?pullRequest=139) <br/>Thank you for your submission, we really appreciate it. Like many open source projects, we ask that you sign our [Contributor License Agreement](https://cla.dev.magento.com/magento/magento-coding-standard?pullRequest=139) before we can accept your contribution.<br/>

lenaorobei:
@piotrekkaminski would be great to have your input here.

fooman:
To add a recent example of a desired change

```
-        return md5(uniqid());
+        return $this->mathRandom->getRandomString(32);
```

the current rule and suggestion would have resulted in something like

```
return hash('sha26', uniqid());
```

orlangur:
Seems like Piotr won't share his opinion on this one :)

@lenaorobei are there some details you could share with us regarding putting on hold?

lenaorobei:
@orlangur this requirement came from the security team and put on hold since there were non resolved concerns.

If possible please create an issue for the future discussion.

jissereitsma:
Would it still be possible to discuss the effectiveness of this rule? I've just scanned the `Magento_Catalog` module for the `Magento2` PHPCS standard and it outputs 8 errors / warnings. And of those, 7 are involving the `md5()` usage. So it seems that either the code needs updating, or the rule needs to be changed, or we need to conclude that the core is not living upto its own rules.

In my opinion, it is best to update the rule. Indeed, `md5()` could lead to security issues. But it doesn't necessary lead to security issues. For instance, the `view/adminhtml/templates/catalog/product/edit/serializer.phtml` is using the `md5()` call simply to randomize an ID to be printed (`md5(microtime())`). Indeed, this could be replaced with `$this->random->getRandomString(10)` but in itself, it is not bad to use `md5()` here. It just pops up as a false positive.

sivaschenko:
@fooman thank you for the pull request. We cannot remove the md5 check from the tests, but you are right about the replacement recommendation, it can be improved.

The check for md5 can be skipped for lines where md5 is used intentionally and there is a good reason why it's preferred over other algorithms. However, a static test should report this issue to ensure it is discussed during code review.

Existing usages of md5 may require review and update.

Issue #147, opened 2019-10-04
PHP files which contain phtml in the name detect as a template


iGerchak:
### Preconditions
1. Magento coding standard 4.0

### Steps to reproduce
1. Create Block with name AmpHtmlBlock.php with the following content:
```php
<?php

namespace Block;

/**
 * Class AmpHtmlBlock
 */
class AmpHtmlBlock
{
    /**
     *
     */
    public function test()
    {
        $this->test();
    }
}
```

### Expected result
There shouldn't be any warnings

### Actual result
```
FILE: .../Block/AmpHtmlBlock.php
-------------------------------------------------------------------------------------------
FOUND 0 ERRORS AND 1 WARNING AFFECTING 1 LINE
-------------------------------------------------------------------------------------------
 15 | WARNING | Usage of $this in template files is deprecated.
-------------------------------------------------------------------------------------------
```

ihor-sviziev:
@lenaorobei could you review this issue? As for me - we need to fix it

lenaorobei:
@iGerchak thanks for reporting. @ihor-sviziev, sure, I agree it needs to be fixed.

PR #152, opened 2019-10-15
#147: Fix for files that contain phtml

larsroettig:
see #147


lenaorobei:
I dont think it is the proper way to fix this issue. If it is PHPCS bug, we need to create one and add a reference to our issue.

larsroettig:
Hi @lenaorobei, 
magento-coding-standard/vendor/squizlabs/php_codesniffer/src/Files/File.php:469 the regex what he uses is the problem.  But I think not so essay to fix it

PR #169, opened 2020-01-09
added @author in forbidden tag

arushibansal013:
Added @author to forbidden tag.
As @author must not be used as per magento documentation


lenaorobei:
Hi @arushibansal013, this change cannot be added to the coding standard, because it is used to check 3rd-party developers extensions, for example on Magento Marketplace, and `@author` is allowed there.

Issue #162, opened 2019-12-11
@deprecated without @see tag false positive


ihor-sviziev:
### Description
Originally this issue appeared here: https://github.com/magento/magento2/pull/21235#issuecomment-564611943

Run static tests against following file:

https://github.com/magento/magento2/blob/5b829a847852f9bcd5ea66d751e0454f33fa3716/app/code/Magento/Customer/Model/Customer.php#L64-L68

Right now you're getting following message:
```
PHP Code Sniffer detected 1 violation(s): 

FILE: /var/www/html/app/code/Magento/Customer/Model/Customer.php
----------------------------------------------------------------------
FOUND 0 ERRORS AND 1 WARNING AFFECTING 1 LINE
----------------------------------------------------------------------
 68 | WARNING | Motivation behind the added @deprecated tag MUST be
    |         | explained. @see tag MUST be used with reference to
    |         | new implementation when code is deprecated and there
    |         | is a new alternative.
----------------------------------------------------------------------



Failed asserting that 1 matches expected 0.
```

As we can see here - there is `@see` tag:
![image](https://user-images.githubusercontent.com/1873745/70642960-d9788a00-1c48-11ea-907d-7cd9173e3c4c.png)

### Expected behavior
You should not have any complains

PR #181, opened 2020-04-04
#162 Modified @deprecated validator to consider it valid if comment is absent but @see tag is set

sinisa86:
Modified @deprecated validator to consider it valid if a comment is absent but @see tag is set.
see #162


Magento2/Helpers/Commenting/PHPDocFormattingValidator.php
```diff
@@ -122,13 +122,18 @@ public function hasDeprecatedWellFormatted($commentStartPtr, $tokens)
...
-            return false;
+            $seeTagRequired = true;
         }
 
         $seePtr = $this->getTagPosition('@see', $commentStartPtr, $tokens);
         if ($seePtr === -1) {
-            return true;
+            if ($seeTagRequired) {
+                return false;
+            } else {
```
lenaorobei, line 134: `else` can be eliminated here.

sinisa86: If we remove "else" branch there, it could lead to unexpected behavior because of the next IF condition:
`
if ($tokens[$seePtr + 2]['code'] !== T_DOC_COMMENT_STRING) {
            return false;
        }
`
That happens because $seePtr could be -1, so $seePtr + 2 points to how-knows-what.
"Else" part is required because it covers behavior when @see is not present at all.

sinisa86: Actually, it really fails on some regular tests when I remove "else" branch.

lenaorobei: `$seeTagRequired` is boolean and if it is `false`, execution will drop to `else` or directly to the `return`.

I propose to simplify code even more:
```
        $seePtr = $this->getTagPosition('@see', $commentStartPtr, $tokens);
        if ($seePtr === -1) {
            return !$seeTagRequired;
        }
        return $tokens[$seePtr + 2]['code'] === T_DOC_COMMENT_STRING;
```

sinisa86:
> @sinisa86 could you please address review comments?

Done.
I'm still a little bit more for an option before applying the last changes because it gives you information about whether `@deprecated` is well-formatted (true/false). Now, when it returns !seeTagRequired, it doesn't tell clearly what's returned in that case (it returns the information that seeTagIsRequired or not, which may seem unrelated to the function name and its purpose). Although this provides the same functionality, it doesn't provide the same level of clarity. 
On the bottom line - it's a just one "IF" condition :)

ihor-sviziev:
@lenaorobei could you check this PR?

ihor-sviziev (approved):

lenaorobei (approved):

lenaorobei:
@sinisa86 thanks for this contribution and your patience!

PR #190, opened 2020-08-17
Run tests on php 7.4

ihor-sviziev:


.travis.yml
```diff
@@ -5,6 +5,7 @@ php:
   - 7.1
   - 7.2
   - 7.3
+  - 7.4
 install: composer install --no-interaction --prefer-source
```
ihor-sviziev, line 9: Btw why we’re  using —prefer-source and not —prefer-dist?

lenaorobei: I believe because we are doing git-based installation. Is there a benefit from using `prefer-dist` option in our context?

ihor-sviziev: Yes, installing from source mean that composer will clone all packages as a fit repo, while “dist” means that packages will be tried to be downloaded as zip files exact commit and just unzipped, usually it much faster

ihor-sviziev: I just created PR with changing to dust, let’s see if it will work

Issue #226, opened 2021-08-24
PHPUnit does not install.


Hexmage:
### Preconditions

1. composer require phpmd/phpmd magento/magento-coding-standard squizlabs/php_codesniffer
2.

### Steps to reproduce

1. composer require phpmd/phpmd magento/magento-coding-standard squizlabs/php_codesniffer
2. run bin/phpcs --config-set installed_paths ~/vendor/magento/magento-coding-standard

### Expected result

1. The command to run

### Actual result

1.
```
 PHP Fatal error:  Uncaught Error: Class 'PHPUnit\Framework\TestCase' not found in ~/vendor/magento/magento-coding-standard/PHP_CodeSniffer/Tests/Standards/AbstractSniffUnitTest.php:23
```

### The Cause
PHPUnit is required as a dev-requirement of magento-coding-standard, which means it will not be installed if you require magento-coding-standard in another composer.json as composer never installs dev-requirements of requirements. So the PHPUnit has to be set as a requirement, seeing as it is now part of the main functionality of the code.

Hexmage:
@magento I am working on this

sivaschenko:
Thanks for reporting the issue and proposed fix, @Hexmage !

I don't think PHPUnit should be a production dependency.

Can you please check if this alternative fix works for you https://github.com/magento/magento-coding-standard/pull/228 ?

Hexmage:
@sivaschenko That should work and is a better solution.

Hexmage:
Sorry about that meant to close my pull request instead of the ticket.

xpoback:
@sivaschenko While PHPUnit should not be a production dependency, its absence should also not break the ability of the script to work in normal mode. I think the fix in #228 is the right solution.
Can you tell when the fix gets included to any tag, so that it can be installed as a "stable" dependency (and not as "dev-master")

PR #227, opened 2021-08-24
Moved PHPUnit from require-dev to require. Solves #226

Hexmage:


sivaschenko:
Thanks for reporting the issue and proposed fix, @Hexmage !

I don't think PHPUnit should be a production dependency.

Can you please check if this alternative fix works for you https://github.com/magento/magento-coding-standard/pull/228 ?

PR #230, opened 2021-08-24
AC-659 Create phpcs static check for ModuleXMLTest

svera:
This sniffer look for occurrences of obsolete attributes `active` and `version` in `module.xml`, and returns a warning if it find any of those.


Magento2/Sniffs/Legacy/ModuleXMLSniff.php
```diff
@@ -0,0 +1,91 @@
...
+        ];
+    }
+
+    /**
+     * @inheritDoc
+     */
+    public function process(File $phpcsFile, $stackPtr)
+    {
+        $line = $phpcsFile->getTokens()[$stackPtr]['content'];
+        if (strpos(trim($line), '<module ') === false) {
```
sivaschenko, line 41: It would be good to validate the file path as well

Magento2/Sniffs/Legacy/ModuleXMLSniff.php
```diff
@@ -0,0 +1,91 @@
...
+            return;
+        }
+        
+        $xml = simplexml_load_string($phpcsFile->getTokensAsString(0, 999999));
```
sivaschenko, line 45: Is that possible to load all token until the token content ending with `/>`?

Also, it would be good to handle possible exceptions for XML operations

Magento2/Sniffs/Legacy/ModuleXMLSniff.php
```diff
@@ -0,0 +1,91 @@
...
+ * Test for obsolete nodes/attributes in the module.xml
+ */
+class ModuleXMLSniff implements Sniff
+{
+    /**
+     * Error violation code.
+     *
+     * @var string
+     */
+    protected $warningCode = 'FoundObsoleteAttribute';
```
sivaschenko, line 23: It's better to use constant here

composer.json
```diff
@@ -10,7 +10,9 @@
     "require": {
         "php": ">=7.3",
         "squizlabs/php_codesniffer": "^3.6",
-        "webonyx/graphql-php": "^14.9"
+        "webonyx/graphql-php": "^14.9",
+        "ext-simplexml": "*",
```
sivaschenko, line 14: Is that possible to avoid additional dependencies by using regex instead of XML queries?

Magento2/Tests/Legacy/ModuleXMLUnitTest.xml
```diff
@@ -0,0 +1,10 @@
+<?xml version="1.0"?>
+<!--
+/**
+ * Copyright © Magento, Inc. All rights reserved.
+ * See COPYING.txt for license details.
+ */
+-->
+<config xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:noNamespaceSchemaLocation="urn:magento:framework:Module/etc/module.xsd">
+    <module name="Magento_TestModule" active="true" version="1" />
```
sivaschenko, line 9: I'd add more test cases
```suggestion
    <module name="Magento_TestModule" active="true" version="1" />
    <module 
        name="Magento_TestModule" 
        active="true" 
        version="1" 
    />
```

Magento2/Sniffs/Legacy/ModuleXMLSniff.php
svera: That's already done in the ruleset

composer.json
svera: Isn't Simplexml bundle in 99,99% of PHP installations? I could try to do it with regexps, but it would be way harder both to create and to read.

Magento2/Tests/Legacy/ModuleXMLUnitTest.xml
svera: I thought about it, but do `module.xml` contain more than a single `module` tag? If you consider it worthwile, I can add this case as well.

Magento2/Sniffs/Legacy/ModuleXMLSniff.php
svera: `getTokenAsString` returns the whole file, as there are no PHP tags inside.

Magento2/Tests/Legacy/ModuleXMLUnitTest.xml
sivaschenko: It would be good to test both structures

Magento2/Sniffs/Legacy/ModuleXMLSniff.php
```diff
@@ -0,0 +1,99 @@
...
+
+/**
+ * Test for obsolete nodes/attributes in the module.xml
+ */
+class ModuleXMLSniff implements Sniff
+{
+    /**
+     * Error violation code.
+     */
+    const WARNING_CODE = 'FoundObsoleteAttribute';
```
sivaschenko, line 21:
```suggestion
    private const WARNING_CODE = 'FoundObsoleteAttribute';
```

sivaschenko:
I think it might be good to take a look at the ways to verify this case using XSD schema, not necessarily with phpcs

sivaschenko:
We have discussed the possible approaches and decided thatt loading XML and verifying specific incorrect nodes/attributes in it is fine. Full XSD validation is still present in dev mode in Magento.

Magento2/ruleset.xml
```diff
@@ -223,6 +223,12 @@
         <severity>8</severity>
         <type>warning</type>
     </rule>
+    <rule ref="Magento2.Legacy.ModuleXML">
+        <include-pattern>*\/module.xml$</include-pattern>
+        <include-pattern>*\/ModuleXMLUnitTest.xml$</include-pattern>
```
sivaschenko, line 228: @svera please remove the test file form the patterns

sivaschenko (changes requested):
@svera can you please take a look at the review comment and update the branch from develop. Thank you!

Magento2/Tests/Legacy/ModuleXMLUnitTest.xml
ihor-sviziev: @svera, it should have two on more module nodes. Please extract it to a separate xml file

ihor-sviziev (changes requested):

sivaschenko:
@ihor-sviziev is it possible to have 2 test files? I don't think we need to follow the exact structure of the module.xml to test a specific case.

ihor-sviziev:
@sivaschenko sure,
Here is an example: 
https://github.com/magento/magento-coding-standard/blob/develop/Magento2/Tests/Classes/DiscouragedDependenciesUnitTest.php
https://github.com/magento/magento-coding-standard/tree/develop/Magento2/Tests/Classes

image

svera:
@ihor-sviziev  could you please clarify this?
> @svera, it should have two on more module nodes. Please extract it to a separate xml file

svera:
> @svera, the regular module.xml file contains the only single section. Having tests with more than 1 section in the file isn't correct as it might cause some issues.
> You can take this test as an example, it uses multiple files :
> https://github.com/magento/magento-coding-standard/blob/develop/Magento2/Tests/Classes/DiscouragedDependenciesUnitTest.php
> 
> https://github.com/magento/magento-coding-standard/tree/develop/Magento2/Tests/Classes
> image

👍 Got it, you're talking about the `<module>` tag, am I wrong?

ihor-sviziev:
yes, for some reason, github removed it from my comment :)

svera:
@ihor-sviziev done

ihor-sviziev (changes requested):
Changes look good now. Could you also add the case, when the module.xml file is valid and we don't have the warning?

svera:
> Changes look good now. Could you also add the case, when the module.xml file is valid and we don't have the warning?

@ihor-sviziev done too

ihor-sviziev (approved):

sivaschenko (approved):

svera:

magento-engcom-team:
@svera the pull request successfully imported.

PR #260, opened 2021-09-06
AC-663: Create phpcs static check for ClassesTest::testPhpCode

svera:
@sivaschenko I'm following what is done in`dev/tests/static/testsuite/Magento/Test/Legacy/ClassesTest.php`, but I'm not sure this is the right approach. What they do is to get all occurrences of certain methods like `getModel()`, `getResourceModel()`, etc. and check the parameters that correspond to a class name. Shouldn't we just check class declarations instead?

Also, I'm not sure about regexps used to check correctness of class names.

### DIFF
--- a/Magento2/Tests/Legacy/ClassesPHPUnitTest.php
+++ b/Magento2/Tests/Legacy/ClassesPHPUnitTest.php
@@ -1,0 +1,30 @@
+<?php
+/**
+ * Copyright © Magento. All rights reserved.
+ * See COPYING.txt for license details.
+ */
+namespace Magento2\Tests\Legacy;
+
+use PHP_CodeSniffer\Tests\Standards\AbstractSniffUnitTest;
+
+class ClassesPHPUnitTest extends AbstractSniffUnitTest
+{
+    /**
+     * @inheritdoc
+     */
+    public function getErrorList()
+    {
+        return [
+            7 => 1,
+            9 => 1,
+        ];
+    }
+
+    /**
+     * @inheritdoc
+     */
+    public function getWarningList()
+    {
+        return [];
+    }
+}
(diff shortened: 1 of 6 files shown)


Magento2/Tests/Legacy/ClassesPHPUnitTest.php
```diff
@@ -0,0 +1,30 @@
...
+class ClassesPHPUnitTest extends AbstractSniffUnitTest
+{
+    /**
+     * @inheritdoc
+     */
+    public function getErrorList()
+    {
+        return [
+            7 => 1,
+            9 => 1,
```
sivaschenko, line 19: Is the expression on line 8 and 10 in the test file perfectly valid in the context of Magento **2**?

jcuerdo (approved):

Magento2/Tests/Legacy/ClassesPHPUnitTest.php
svera: @sivaschenko what are perfectly valid expression in the context of Magento 2? I've made up those examples as an interpretation of the regular expression, but as you said this may be no longer accurate. Sniff in particular refers to PSR1 https://www.php-fig.org/psr/psr-1.

sivaschenko: I believe we should simplify the test to fail in case of any static calls to these methods independently from the passed parameter. Also, I'd fail the test if `Mage` class is mentioned

sivaschenko:
We have discussed the test with @svera and come to the next conclusion:

The test should be replaced with 2 separate checks

- Sniff checking the legacy Magento 1 code in the codebase by looking for `Mage` class and related static methods calls (see the PR code)
- Sniff checking the namespace and class name are matching the PSR1 (need to check if this test is already present in our ruleset and if it is provided by phpcs out of the box)

svera:
* Changed test to look for `Mage` class static methods usages (although I'm not sure abut the usefulness of this test).
* PSR1 checks are already present in `ruleset.xml` (`Squiz.Classes.ValidClassName` and `PSR1.Classes.ClassDeclaration`)

svera:
Closing PR as the case covered here is already covered by `\Magento2\Sniffs\Legacy\MageEntitySniff`

PR #270, opened 2021-09-13
Fixed wrongly returning error for valid descriptions

svera:
#140 was wrongly returning error for valid short descriptions. We added several unit tests checking for this issue, as well as not allowing to repeat property class name in short description and rephrasing the error return when a short description is placed after a @var annotation.

### DIFF
--- a/Magento2/Sniffs/Commenting/ClassPropertyPHPDocFormattingSniff.php
+++ b/Magento2/Sniffs/Commenting/ClassPropertyPHPDocFormattingSniff.php
@@ -143,8 +143,9 @@ class ClassPropertyPHPDocFormattingSniff extends AbstractVariableSniff
             return;
         };
         $propertyName = trim($tokens[$propertyNamePosition]['content'], '$');
+        $shortDescription = strtolower($tokens[$isShortDescriptionPreviousVar]['content']);
 
-        if (strtolower($tokens[$isShortDescriptionPreviousVar]['content']) === strtolower($propertyName)) {
+        if ($shortDescription === strtolower($propertyName)) {
             $error = 'Short description duplicates class property name.';
             $phpcsFile->addWarning($error, $isShortDescriptionPreviousVar, 'AlreadyHaveMeaningfulNameVar');
             return;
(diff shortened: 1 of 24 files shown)


Magento2/Sniffs/Commenting/ClassPropertyPHPDocFormattingSniff.php
```diff
@@ -125,23 +126,28 @@ public function processMemberVar(File $phpcsFile, $stackPtr)
...
-            preg_match(
-                '`^((?:\|?(?:array\([^\)]*\)|[\\\\\[\]]+))*)( .*)?`i',
-                $tokens[($foundVar + 2)]['content'],
-                $varParts
-            );
-            if ($varParts[1]) {
-                return;
-            }
+
+        if (stripos($tokens[$isShortDescriptionPreviousVar]['content'], $tokens[$string]['content']) !== false) {
```
sivaschenko, line 130: Please check that $isShortDescriptionPreviousVar is not false before using it as an array key

Magento2/Sniffs/Commenting/ClassPropertyPHPDocFormattingSniff.php
```diff
@@ -125,23 +126,28 @@ public function processMemberVar(File $phpcsFile, $stackPtr)
...
+                | (?<=[A-Z])      # Or g2of2; Position is after uppercase,
+                  (?=[A-Z][a-z])  # and before upper-then-lower case.
+                /x';
+        $varTagParts = preg_split($re, $tokens[$string]['content']);
+
+        foreach ($varTagParts as $part) {
+            if (stripos($tokens[$isShortDescriptionPreviousVar]['content'], $part) === false) {
+                return;
+            }
+        }
```
sivaschenko, line 148: I don't think finding a part of property name in the description should fail the test, I'd rather check if the short description matches the split variable name.
```suggestion
            if (stripos($tokens[$isShortDescriptionPreviousVar]['content'], implode(' ', $varTagParts)) === false) {
                return;
            }
```

sivaschenko (changes requested):
Thank you for the fix @svera ! Please see my comments

sivaschenko (approved):

svera:

PR #324, opened 2021-11-04
Excluded JS files for PSR sniffs

sivaschenko:
Excluded JS files for PSR sniffs and the indents requirements might conflict with eslint static tests


sivaschenko:

m2-github-services:

PR #332, opened 2021-11-15
AC-1059: Move PHPCompatibility rules from Magento tests

svera:
Declare PHPCompatibility as a dependency of magento-coding-standard
Include `/dev/tests/static/testsuite/Magento/Test/Php/_files/PHPCompatibilityMagento/ruleset.xml` in the `Magento2/ruleset.xml`


PHPCompatibilityMagento/ruleset.xml
```diff
@@ -0,0 +1,16 @@
...
+/**
+ * Copyright © Magento, Inc. All rights reserved.
+ * See COPYING.txt for license details.
+ */
+-->
+<ruleset name="PHPCompatibilityMagento">
+    <description>Magento 2 specific ruleset which checks for PHP cross version compatibility.</description>
+    <rule ref="../vendor/phpcompatibility/php-compatibility/PHPCompatibility">
+        <exclude name="PHPCompatibility.Miscellaneous.RemovedAlternativePHPTags.MaybeASPOpenTagFound" />
+        <exclude name="PHPCompatibility.FunctionUse.RequiredToOptionalFunctionParameters" />
```
svera, line 12: This sniff had to be excluded because they return a `PHP Fatal error:  Cannot declare class PHPCompatibility\Sniffs\FunctionUse\RequiredToOptionalFunctionParametersSniff, because the name is already in use` error. The same happens with the following 2 exclusions. Ideas are welcomed.

sivaschenko: @svera can you please check how these names are allocated initially.

Also, can you please add the compatibility rule to Magento2 standard ruleset.

svera: @sivaschenko it seems the fatal error mentioned above is a known issue

https://github.com/PHPCompatibility/PHPCompatibility/issues/793

unfortunately, the latest release was published before the fix of this issue. What do you think we should do here?

svera: @sivaschenko  I finally found the issue. PHPCompatibility was not registered in PHPCS, thus the only way to include it was using its path in the `vendor` folder which, in turn, caused a bug in PHPCompatibility to raise up. Solution is easy, just register PHPCompatibility as an installed path of PHPCS through `composer.json`

svera:
@sivaschenko static tests check OK: https://github.com/magento-commerce/magento2ce/pull/7224

eliseacornejo (dismissed):

sivaschenko (dismissed):

sivaschenko:

m2-github-services:

svera:

m2-github-services:
@svera an error occurred during the Pull Request import.

m2-github-services:
@svera an error occurred during the Pull Request import.

sivaschenko (approved):

sivaschenko:

m2-github-services:

PR #335, opened 2021-11-19
Fix Comment block is missing Failure when Attribute is present after the docblock

karyna-tsymbal-atwix:
The Warning 'Comment block is missing' is thrown when the attribute is present after the docblock.

example:
```
/**
 * Check if given type is any type.
 *
 * @param string $type
 * @return bool
 */
#[\ReturnTypeWillChange]
public function isTypeAny($type)
```

In such case our Sniff should ignore the attribute.

### DIFF
--- a/Magento2/Sniffs/Annotation/MethodArgumentsSniff.php
+++ b/Magento2/Sniffs/Annotation/MethodArgumentsSniff.php
@@ -70,7 +70,23 @@ class MethodArgumentsSniff implements Sniff
     private function validateCommentBlockExists(File $phpcsFile, int $previousCommentClosePtr, int $stackPtr): bool
     {
         $tokens = $phpcsFile->getTokens();
+        $attributeFlag = false;
         for ($tempPtr = $previousCommentClosePtr + 1; $tempPtr < $stackPtr; $tempPtr++) {
+            $tokenCode = $tokens[$tempPtr]['code'];
+
+            // Ignore attributes e.g. #[\ReturnTypeWillChange]
+            if ($tokenCode === T_ATTRIBUTE_END) {
+                $attributeFlag = false;
+                continue;
+            }
+            if ($attributeFlag) {
+                continue;
+            }
+            if ($tokenCode === T_ATTRIBUTE) {
+                $attributeFlag = true;
+                continue;
+            }
+
             if (!$this->isTokenBeforeClosingCommentTagValid($tokens[$tempPtr]['type'])) {
                 return false;
             }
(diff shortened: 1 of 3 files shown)


Magento2/Sniffs/Annotation/MethodArgumentsSniff.php
```diff
@@ -70,7 +70,23 @@ private function isTokenBeforeClosingCommentTagValid(string $type): bool
     private function validateCommentBlockExists(File $phpcsFile, int $previousCommentClosePtr, int $stackPtr): bool
     {
         $tokens = $phpcsFile->getTokens();
+        $attributeFlag = false;
         for ($tempPtr = $previousCommentClosePtr + 1; $tempPtr < $stackPtr; $tempPtr++) {
+            $tokenCode = $tokens[$tempPtr]['code'];
+
+            // Ignore attributes e.g. #[\ReturnTypeWillChange]
```
sivaschenko, line 77: What do you think about just adding T_ATTRIBUTE (and other attribute related tokens) to `validTokensBeforeClosingCommentTag` property? (would be good to actually change this property to const)

I believe the idea of the `validTokensBeforeClosingCommentTag` token codes is actually to list tokens allowed between comment and function - sounds like exactly what we need.

sivaschenko (changes requested):
Thanks for the PR @karyna-tsymbal-atwix ! Please see my comment.

Also, please add several different cases with valid and invalid phpdoc to the unit tests

karyna-tsymbal-atwix (commented):
Hello @karyna-tsymbal-atwix 
Thank you for your contribution here
Approved ✔️

andrewbess (dismissed):
Hello @karyna-tsymbal-atwix 
Thank you for your contribution here
Approved ✔️

Magento2/Sniffs/Annotation/MethodArgumentsSniff.php
karyna-tsymbal-atwix: Hi @sivaschenko , as we don't have a single token for the attribute, in that case we need to add T_ATTRIBUTE, T_ATTRIBUTE_END, T_NS_SEPARATOR, T_STRING  and other tokens that possibly can be in other attributes in the future,
e.g. something like
```
#[ListensTo(ProductCreated::class)]
#[\Attribute(\Attribute::TARGET_METHOD | \Attribute::TARGET_PROPERTY)]
```

So I think it's better just to ignore everything between #[ and ]

karyna-tsymbal-atwix:
@sivaschenko thanks for the review! I've added couple test cases, please check.

andrewbess (approved):

sivaschenko (approved):

sivaschenko:

m2-github-services:

karyna-tsymbal-atwix:
This PR contains changens according to https://github.com/magento/magento-coding-standard/pull/335

PR #341, opened 2021-12-13
AC-1717: Passing null to non-nullable internal function parameters is…

svera:
Passing null to non-nullable internal function parameters is deprecated:

* `preg_split` --> In this case the parameter `$limit` cannot be `null` anymore, it has to be replaced by `-1` (that means in both cases without limit)
* `DateTime` --> In this case `DateTime::__construct(string $datetime = "now"`... parameter `$datetime` cannot be null anymore. We need to replace it with `"now"`
* `mb_strpos()` --> In this case the parameter `$offset` cannot be `null`. We need to replace with `0`


eliseacornejo (commented):
Can you add coypright in tests files too, please?

loginesta (approved):

eliseacornejo (approved):

svera:

m2-github-services:
@svera the Pull Request is successfully imported.

PR #345, opened 2021-12-16
AC-1718 return types for arrayaccess methods

svera:
Since PHP 8.1 most of the PHP internal classes are updated with a tentative return type. If you implement these classes you need to add the return type.
This rule adds a return type if it is not there when a class implements the `ArrayAccess` interface.


loginesta (approved):

svera:

m2-github-services:
@svera the Pull Request is successfully imported.

PR #362, opened 2022-01-14
AC-1895: Avoid having duplicated issues in magento-coding-standard

svera:
There are a couple of sniffs duplicated between PHPCS and PHP Compatibility. I've picked the ones with better coverage after doing some tests.

| Excluded | Kept | Description |
| --- | --- | --- |
| Generic.PHP.DeprecatedFunctions | PHPCompatibility.FunctionUse.RemovedFunctions | Excluded one only catches deprecations and, due to it using `get_defined_functions()` to get deprecated methods, it doesn't include functions coming from extensions or deprecations from different PHP versions than the one which is running the sniff. |
| PHPCompatibility.Syntax.ForbiddenCallTimePassByReference | Generic.Functions.CallTimePassByReference | Both are equivalent, there is no special advantage of one over the other; in fact, call-time pass-by-reference is forbidden since PHP 5.4 so we can remove both checks. |

### DIFF
--- a/Magento2/ruleset.xml
+++ b/Magento2/ruleset.xml
@@ -164,6 +164,10 @@
         <severity>10</severity>
         <type>error</type>
     </rule>
+    <rule ref="PHPCompatibility.FunctionUse.RemovedFunctions">
+        <severity>10</severity>
+        <type>error</type>
+    </rule>
     <!-- Severity 9 warnings: Possible security and issues that may cause bugs. -->
     <rule ref="Generic.Files.ByteOrderMark">
         <severity>9</severity>
@@ -761,8 +765,7 @@
         <exclude name="PHPCompatibility.Constants.NewMagicClassConstant.Found" />
         <exclude name="PHPCompatibility.Classes.ForbiddenAbstractPrivateMethods.Found" />
         <!-- Following sniffs have an equivalent in PHPCS -->
-        <exclude name="PHPCompatibility.Syntax.ForbiddenCallTimePassByReference.Deprecated" />
-        <exclude name="PHPCompatibility.Syntax.ForbiddenCallTimePassByReference.NotAllowed" />
+        <exclude name="PHPCompatibility.Syntax.ForbiddenCallTimePassByReference" />
         <!-- Check for cross-version support for stated PHP version and higher. -->
         <config name="testVersion" value="7.4-"/>
     </rule>
--- a/Magento2/ruleset.xml
+++ b/Magento2/ruleset.xml
@@ -16,11 +16,6 @@
         <exclude-pattern>*\.phtml$</exclude-pattern>
         <exclude-pattern>*\.xml$</exclude-pattern>
     </rule>
-    <rule ref="Generic.PHP.DeprecatedFunctions">
-        <severity>10</severity>
-        <type>error</type>
-        <exclude-pattern>*\.xml$</exclude-pattern>
-    </rule>
     <rule ref="Generic.PHP.NoSilencedErrors">
         <severity>10</severity>
         <type>error</type>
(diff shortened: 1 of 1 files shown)


Magento2/ruleset.xml
```diff
@@ -16,11 +16,6 @@
         <exclude-pattern>*\.phtml$</exclude-pattern>
         <exclude-pattern>*\.xml$</exclude-pattern>
     </rule>
-    <rule ref="Generic.PHP.DeprecatedFunctions">
```
svera, line 19: This only catches deprecations and, due to it using `get_defined_functions()`to get deprecated methods, its coverage is not as good as the solution from PHPCompatibility because it doesn't include extensions coming from functions or deprecations from different PHP versions than the one which is running the sniff.

sivaschenko (changes requested):
@svera can you please add information describing which sniff was excluded/removed in favour to which sniff, what are the difference between them and pros/cons. This should help to understand this decision better. Thank you!

svera:
@sivaschenko adressed in PR description

sivaschenko:
Grouped excluded `PHPCompatibility.Syntax.ForbiddenCallTimePassByReference` and set correct error 10 severity for `PHPCompatibility.FunctionUse.RemovedFunctions`

https://github.com/magento/magento-coding-standard/pull/362/commits/72be356f4e5cf07dc553fd5a0768763601cd7fdf

sivaschenko (approved):

sivaschenko:

m2-github-services:

tuyennn:
@sivaschenko any idea while I still got
```
phpcs: ERROR: Referenced sniff "PHPCompatibility.FunctionUse.RemovedFunctions" does not exist 
```
On my IDE

rfortes-scrubsandbeyond:
@tuyennn I'm also having this issue. Did you manage to resolve this?

rafafortes:
@tuyennn I'm also having this issue. Did you manage to resolve this?

sivaschenko:
@tuyennn @rafafortes I believe you need to register PHPCompatibility standard:

For example from magento project root `vendor/bin/phpcs --config-set installed_paths ../../..,../../phpcompatibility/php-compatibility/PHPCompatibility` (or adjust the paths according to working directory)

That is handled for Magento by `post-install-cmd` and `post-update-cmd` in `composer.json`, but may not be automatically handled for other specific installations

rafafortes:
@sivaschenko it resolved the issue through the CLI but PHPStorm still complains about it, that's odd.

tuyennn:
@rafafortes @sivaschenko It is really annoying me, so I decided to revert to the older tag version of [magento-coding-standard](https://github.com/magento/magento-coding-standard)

tuyennn:
@rafafortes you could find the workaround here https://github.com/PHPCompatibility/PHPCompatibility/issues/459#issuecomment-622006830

Vinai:
I'm running into the same issue @rafafortes.  
PHPCompatibility is properly registered in `vendor/squizlabs/php_codesniffer/CodeSniffer.conf`:

```php
<?php
 $phpCodeSnifferConfig = array (
  'installed_paths' => '../../magento/magento-coding-standard,../../phpcompatibility/php-compatibility',
)
?>
```

I've checked that PHPStorm is indeed running the sniff from the Magento base directory as the current working directory (just like when running it from the command line).

Do you have any other ideas what might be causing the difference between running `phpcs` from the CLI and from PHPStorm?

rafafortes:
@Vinai  I've tried things like adding more installed_paths and different phpcs binaries (from OS, Docker containers) and none of this resolved the issue so far. The PHPStorm debug messages don't help also so I'm running out of ideas

tuyennn:
@rafafortes @Vinai I have to set the **Installed standard path** of Magento Coding Stardard to `vendor/magento/magento-coding-standard` for PHPStorm
image

Also need to set installed_paths PHPCompatibility on root of the project as
```
vendor/bin/phpcs --config-set installed_paths  ../phpcompatibility/php-compatibility/PHPCompatibility/
```
To avoid annoying error with PHP CodeSniffer, the main problem is I have CodeSniffer Magento Coding Stardard ran at deprecated version such as 6 instead 21

tuyennn:
@rafafortes here how I resolved the annoying PHPStorm errors with new magento codding standard. 

1 . From project I remove the current require-dev magento codding standard. We aim to use the external magento codding standard outside project because the one from vendor was always deprecated, since we cannot keep the one from vendor run along with external up-to-date magento codding standard source
```
composer remove magento/magento-coding-standard
```
2. Uncheck **Installed standard path** from PHPStorm and from the root of current project do
```
 vendor/bin/phpcs --config-set installed_paths  your_local_external_path_to/magento-coding-standard,../../phpcompatibility/php-compatibility
```

For example
```
 vendor/bin/phpcs --config-set installed_paths ~/Projects/magento-coding-standard,../../phpcompatibility/php-compatibility
 ```

Vinai:
@rafafortes To debug I've added a `require 'debug.php';` to vendor/squizlabs/php_codesniffer/CodeSniffer.conf`.
Even though the file ends with `.conf`, it's loaded with `include` or `require`.
Then I added a ´debug.php` and started to add debug statements and breakpoints.

Vinai:
I don't remember. Will have to check again. Using a PHP interpreter config with XDEBUG and starting the debug listener in PHPStorm should work, but I am not sure.
If that doesn't work, using var_dump'n'die and debug_backtrace is always an oldschool option I guess :)

rafafortes:
Thank you @Vinai . The old and good var_dump helped me out!

I realized the $resolvedInstalledPaths array was missing the path: '/opt/project/vendor/phpcompatibility/php-compatibility/PHPCompatibility';

As a quick fix I ended up adding the missing path manually to the array as follows:

// vendor/squizlabs/php_codesniffer/src/Util/Standards.php
$resolvedInstalledPaths[] = '/opt/project/vendor/phpcompatibility/php-compatibility/PHPCompatibility';
return $resolvedInstalledPaths;

PR #374, opened 2022-02-14
AC-2219: Magento2.Functions.FunctionsDeprecatedWithoutArgument

loginesta:
https://jira.corp.magento.com/browse/AC-2219 Magento2.Functions.FunctionsDeprecatedWithoutArgument

### DIFF
--- a/Magento2/Sniffs/Functions/FunctionsDeprecatedWithoutArgumentSniff.php
+++ b/Magento2/Sniffs/Functions/FunctionsDeprecatedWithoutArgumentSniff.php
@@ -36,7 +36,7 @@ class FunctionsDeprecatedWithoutArgumentSniff implements Sniff
      * @var array
      */
     private const DEPRECATED_FUNCTIONS_AND_FIXES = [
-        'mb_check_encoding' => 'null',
+        'mb_check_encoding' => '\'\'',
         'get_class' => '$this',
         'get_parent_class' => '$this'
     ];
--- a/Magento2/Sniffs/Functions/FunctionsDeprecatedWithoutArgumentSniff.php
+++ b/Magento2/Sniffs/Functions/FunctionsDeprecatedWithoutArgumentSniff.php
@@ -77,12 +77,14 @@ class FunctionsDeprecatedWithoutArgumentSniff implements Sniff
             );
             return;
         }
+        
+        $fix = $phpcsFile->addFixableWarning(
+            sprintf(self::WARNING_MESSAGE, $functionName),
+            $stackPtr,
+            self::WARNING_CODE
+        );
 
-        if ($phpcsFile->addFixableWarning(
-                sprintf(self::WARNING_MESSAGE, $functionName),
-                $stackPtr,
-                self::WARNING_CODE
-            ) === true) {
+        if ($fix === true) {
             $content = self::DEPRECATED_FUNCTIONS_AND_FIXES[$functionName];
             $phpcsFile->fixer->beginChangeset();
             $phpcsFile->fixer->addContentBefore($phpcsFile->findNext(T_CLOSE_PARENTHESIS, $stackPtr), $content);
--- a/Magento2/Sniffs/Functions/FunctionsDeprecatedWithoutArgumentSniff.php
+++ b/Magento2/Sniffs/Functions/FunctionsDeprecatedWithoutArgumentSniff.php
@@ -36,7 +36,7 @@ class FunctionsDeprecatedWithoutArgumentSniff implements Sniff
      * @var array
      */
     private const DEPRECATED_FUNCTIONS_AND_FIXES = [
-        'mb_check_encoding' => '[]',
+        'mb_check_encoding' => 'STDIN',
         'get_class' => '$this',
         'get_parent_class' => '$this'
     ];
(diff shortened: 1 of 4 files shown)


Magento2/Sniffs/Functions/FunctionsDeprecatedWithoutArgumentSniff.php
```diff
@@ -31,15 +31,15 @@ class FunctionsDeprecatedWithoutArgumentSniff implements Sniff
...
-    private const FUNCTIONS_LIST = [
-        'mb_check_encoding',
-        'get_class',
-        'get_parent_class',
-        'get_called_class'
+    private const DEPRECATED_FUNCTIONS_AND_FIXES = [
+        'mb_check_encoding' => 'null',
+        'get_class' => '$this',
+        'get_parent_class' => '$this',
+        'get_called_class' => '$this'
```
sivaschenko, line 42: Using arguments with get_called_class leads to the following error: `Fatal error: Uncaught ArgumentCountError: get_called_class() expects exactly 0 arguments, 1 given`
Please remove this function from the list

Magento2/Sniffs/Functions/FunctionsDeprecatedWithoutArgumentSniff.php
```diff
@@ -31,15 +31,15 @@ class FunctionsDeprecatedWithoutArgumentSniff implements Sniff
     private const WARNING_CODE = 'FunctionsDeprecatedWithoutArgument';
 
     /**
-     * Deprecated functions without argument.
+     * Deprecated functions without argument
```
sivaschenko, line 34: Let's specify the source of data
```suggestion
     * Deprecated functions without argument https://wiki.php.net/rfc/deprecations_php_8_1
```

Magento2/Sniffs/Functions/FunctionsDeprecatedWithoutArgumentSniff.php
```diff
@@ -31,15 +31,15 @@ class FunctionsDeprecatedWithoutArgumentSniff implements Sniff
...
      *
      * @var array
      */
-    private const FUNCTIONS_LIST = [
-        'mb_check_encoding',
-        'get_class',
-        'get_parent_class',
-        'get_called_class'
+    private const DEPRECATED_FUNCTIONS_AND_FIXES = [
+        'mb_check_encoding' => 'null',
```
sivaschenko, line 39: `mb_check_encoding(null);` results in the following error `Deprecated: mb_check_encoding(): Calling mb_check_encoding() without argument is deprecated in /in/HGVsB on line 8` please correct the argument

sivaschenko (changes requested):
Hi @loginesta thanks for PR, please check my comments and test the proposed autofixed code.

Magento2/Sniffs/Functions/FunctionsDeprecatedWithoutArgumentSniff.php
loginesta: That's exactly what I was suspecting...

Magento2/Sniffs/Functions/FunctionsDeprecatedWithoutArgumentSniff.php
loginesta: check the latest commits, please

Magento2/Sniffs/Functions/FunctionsDeprecatedWithoutArgumentSniff.php
```diff
@@ -31,15 +31,14 @@ class FunctionsDeprecatedWithoutArgumentSniff implements Sniff
...
      *
      * @var array
      */
-    private const FUNCTIONS_LIST = [
-        'mb_check_encoding',
-        'get_class',
-        'get_parent_class',
-        'get_called_class'
+    private const DEPRECATED_FUNCTIONS_AND_FIXES = [
+        'mb_check_encoding' => '\'\'',
```
sivaschenko, line 39: Is `mb_check_encoding('')` equivalent to `mb_check_encoding()`?

sivaschenko:
The negative scenario with `mb_check_encoding` function can be tested using, for example, the following code fragment:

```
$correctText = 'アプリ';
$incorrectText = substr($correctText, 0, strlen($correctText) - 1); // mb_strlen() should be used instead
var_dump(mb_check_encoding($incorrectText)); // false
```

I've checked mb_check_encoding without arguments passing incorrect encoding value as PHP CLI argument and GET parameter. In both cases `mb_check_encoding()` returned `true`.

Also, I was not able to identify any specific default value for the first argument of the function defined in the [php-src](https://github.com/php/php-src/blob/dfd2a80b69d469d879d60c12c1d9d9ca49cb3f76/ext/mbstring/mbstring.c#L3924)

I don't think the function works as described in the [documentation](https://www.php.net/manual/en/function.mb-check-encoding.php). It does not validate "all the input". In this case there is no specific autofixing suggestion for this function. Autofixing for it will not be provided

eliseacornejo (approved):

sivaschenko:

m2-github-services:

PR #387, opened 2022-04-19
Allow empty constructor function blocks when using property promotion

### LINKED ISSUE #385: Constructor property promotion and EmptyBlock rule
### Description

Using the new PHP 8.0 feature of constructor property promotion, the constructor function will have an empty block:
```php
public function __construct(
    private RequestInterface $request
) {}
```
This is not allowed with the current ruleset, as it violates the Magento2.CodeAnalysis.EmptyBlock rule:
```
```

mfickers:
This fixes #385 

This PR introduces an additional exception to the EmptyBlock rule. Empty constructor function blocks will no longer result in errors in case there is at least one parameter and every parameter has a visibility set.

This improves the coding standard, as it allows the use of the new PHP 8.0 feature of property promotion. For constructors where every single parameter is promoted, an empty function block is expected and does not represent a code smell.

### DIFF
--- a/Magento2/Sniffs/CodeAnalysis/EmptyBlockSniff.php
+++ b/Magento2/Sniffs/CodeAnalysis/EmptyBlockSniff.php
@@ -37,6 +37,17 @@ class EmptyBlockSniff extends EmptyStatementSniff
             return;
         }
 
+        // Ignore empty constructor function blocks when using property promotion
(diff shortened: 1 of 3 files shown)


sivaschenko (approved):

sivaschenko:

m2-github-services:

PR #400, opened 2022-06-02
Check @deprecated format only if file has comment

### LINKED ISSUE #399: Calling hasDeprecatedWellFormatted with a false $commentStartPtr leads to error
### Preconditions

1. Magento Coding Standard containing https://github.com/magento/magento-coding-standard/commit/e64106ddbdb9ee91ce89bb02b7858082eea21cf8.

### Steps to reproduce

1. Run `phpcs --standard=Magento2 Foo.php` for the following sample file

```php
<?php

namespace Foo\Bar;

class Foo
{

    protected function foo()
    {
        return 1;
    }
}
```

### Expected result

1. phpcs complains about a missing comment block.

### Actual result

1. A sniff fails with an error message:

> An error occurred during processing; checking has been aborted. The error message was: Undefined index: comment_closer in /path/to/magento-coding-standard/Magento2/Helpers/Commenting/PHPDocFormattingValidator.php on line 142

Alternatively, in some situations the phpcs process hangs in an eternal loop. (Due to ignoring the error and going into the following `for` loop with bad limits.)

mattijv:
The MethodAnnotationStructureSniff throws an error if the file does
not have a comment ($commentStartPtr is false). This commit moves the
@deprecated format check after we check that the file has a comment.

This fixes #399.

### DIFF
--- a/Magento2/Sniffs/Annotation/MethodAnnotationStructureSniff.php
+++ b/Magento2/Sniffs/Annotation/MethodAnnotationStructureSniff.php
@@ -52,6 +52,11 @@ class MethodAnnotationStructureSniff implements Sniff
     {
         $tokens = $phpcsFile->getTokens();
         $commentStartPtr = $phpcsFile->findPrevious(T_DOC_COMMENT_OPEN_TAG, ($stackPtr), 0);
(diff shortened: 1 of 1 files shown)


sivaschenko:

m2-github-services:

PR #406, opened 2022-06-24
Class properties that have types should not require a DocBlock

### LINKED ISSUE #404: Do not require DocBlock for typed class attributes
### Expected behavior
Class attributes that have specified types should not require a DocBlock. When an attribute already has a type, the DocBlock is providing no new information, and is instead just creating noise - e.g.:
```
/** @var ProductRepositoryInterface */
private ProductRepositoryInterface $productRepository;
We already have the type information - we don't need the DocBlock to tell us the type.
```

I propose that the rule regarding class attributes requiring a `@var` annotation with a type be changed so that it only applies to attributes without a specified type.

### Benefits
Removes the need for redundant comments

aligent-lturner:
Fixes #404 

If a class property has a type, then there should not be a requirement for a DocBlock.
A DocBlock can still be added if desired - this change simply removes the warning for a missing comment block when the property has a defined type.

Note - this does not validate that the type is valid, only that it is not specifically _**invalid**_ in the same way as https://github.com/magento/magento-coding-standard/blob/develop/Magento2/Sniffs/Annotation/MethodArgumentsSniff.php

Additional:  This issue was originally raised against the documentation page (https://github.com/AdobeDocs/commerce-php/issues/23), and was suggested to be raised directly here. The documentation would need to be updated to reflect that DocBlocks are only required for properties without a type.

### DIFF
--- a/Magento2/Tests/Commenting/ClassPropertyPHPDocFormattingUnitTest.inc
+++ b/Magento2/Tests/Commenting/ClassPropertyPHPDocFormattingUnitTest.inc
@@ -194,4 +194,6 @@ class correctlyFormattedClassMemberDocBlock
      * @see Message with some reference
      */
     protected string $itIsCorrect;
+
+    private string $typedPropertyWithoutComment;
 }
--- a/Magento2/Sniffs/Commenting/ClassPropertyPHPDocFormattingSniff.php
+++ b/Magento2/Sniffs/Commenting/ClassPropertyPHPDocFormattingSniff.php
@@ -38,6 +38,16 @@ class ClassPropertyPHPDocFormattingSniff extends AbstractVariableSniff
     private $PHPDocFormattingValidator;
 
     /**
+     * @var array
+     */
+    private $invalidTypes = [
+        'null',
+        'false',
+        'true',
+        'self'
+    ];
+
+    /**
      * Constructs an ClassPropertyPHPDocFormattingSniff.
      */
     public function __construct()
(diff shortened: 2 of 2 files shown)


aligent-lturner:
Failing tests do not appear to be related to this change, but a missing Symfony class:

```
Error: Class "Symfony\Component\DependencyInjection\Loader\Configurator\ContainerConfigurator" not found
```

Magento2/Tests/Commenting/ClassPropertyPHPDocFormattingUnitTest.inc
```diff
@@ -194,4 +194,6 @@ class correctlyFormattedClassMemberDocBlock
      * @see Message with some reference
      */
     protected string $itIsCorrect;
+
+    private string $typedPropertyWithoutComment;
```
ihor-sviziev, line 198: I would add more use cases, like there https://3v4l.org/MLIOe:

```suggestion

    private string $typedPropertyWithoutComment;

    private string $typedPropertyWithoutComment2;

    private ?string $typedPropertyWithoutComment3;

    private string|int $typedPropertyWithoutComment4;

    private Test $typedPropertyWithoutComment5;

    private Test|Test2 $typedPropertyWithoutComment6;
    
    private ?Test $typedPropertyWithoutComment7;    
```

ihor-sviziev (changes requested):

Magento2/Tests/Commenting/ClassPropertyPHPDocFormattingUnitTest.inc
```diff
@@ -194,4 +194,19 @@ class correctlyFormattedClassMemberDocBlock
      * @see Message with some reference
      */
     protected string $itIsCorrect;
+
+
```
ihor-sviziev, line 198: Oops. I did a mistake, there should be one line, not two

```suggestion

```

torhoehn:
@sivaschenko Would it be possible to merge this please?

ihor-sviziev (changes requested):
hi @aligent-lturner,
Could you please resolve conflicts?

aligent-lturner:
> hi @aligent-lturner, Could you please resolve conflicts?

@ihor-sviziev conflict has been resolved

ihor-sviziev (approved):

sinhaparul:
@magento import PR to magento-commerce/magento-coding-standard repository

m2-github-services:
@sinhaparul the Pull Request is successfully imported.

fredden:
See also #476

ihor-sviziev (approved):

ihor-sviziev:
@sinhaparul @sidolov could you please merge it?

aligent-lturner:
@sidolov this has been approved for almost a year now. Any chance it can be merged soon?

PR #429, opened 2022-11-30
AC-7098: Performance issue in TableNameSniff with big array

victor-v-rad:
### Fix summary

The performance of TableNameSniff degradated with a big array and structure like this
`$fooBar=array(0=>array(125,0,875,750),1 =>array(85,0,194,716), 2 => 100);` 
But TableNameSniff expected to process the associative array similar to
`['main_table_name' => 'magento/sample_table']` 
in result, [$phpcsFile->findPrevious](https://github.com/magento/magento-coding-standard/blob/1452ed3dd49f11b2b7743048595e19e02e45f402/Magento2/Sniffs/Legacy/TableNameSniff.php#L191-L194) could not find associate array key in [$stackPtr -1](https://github.com/magento/magento-coding-standard/blob/1452ed3dd49f11b2b7743048595e19e02e45f402/Magento2/Sniffs/Legacy/TableNameSniff.php#L193)

**The fix is to make findPrevious to find any non whitespace token as array key** 

### Related Pull Requests
 https://github.com/magento-commerce/mray/pull/158

PR #430, opened 2022-11-30
autotests fail on mainbranch

victor-v-rad:
autotests fail on mainbranch


victor-v-rad:
Trying to fix here failed https://github.com/magento/magento-coding-standard/pull/429

PR #436, opened 2023-02-01
Simplify installation instructions

### LINKED ISSUE #390: PHPStorm unable to run phpcs inspection
### Preconditions
1. Tested on Magento Open Source 2.4.4 and Adobe Commerce 2.4.4
2. PHP 8.1 (I haven't tried 7.4)
3. PHPStorm version is 2021.2.4

### Steps to reproduce
1. Configure PHP interpreter in PHPStorm
2. Configure phpcs in PHPStorm to use the Magento coding standard
3. Enable the phpcs inspections

### Expected result
1. PHPStorm displays phpcs rule violations according to the configuration

### Actual result
In the PHStorm event log the message is shown
```
ERROR: Referenced sniff "PHPCompatibility.FunctionUse.RemovedFunctions" does not exist
```

### Additional Information

Running phpcs on the command line works
```
vendor/bin/phpcs --standard=vendor/magento/magento-coding-standard/Magento2 path/to/check
```

---

The same PHPStorm configuration works on Magento 2.4.3 with PHP 7.4. 

---

If I remove the reference to that sniff from the `Magento2/ruleset.xml`, phpcs fails with a stacktrace where any other Magento2 rule extending from a `PHPCompatibility` rule class can't find the parent class.   
Maybe this points to an autoload error?
However, I quick test by adding `"PHPCompatibility\\:": "vendor/phpcompatibility/php-compatibility/PHPCompatibility/"` to the root composer.json and dumping the composer autoloader did not change anything.

fredden:
The installation instructions for this coding standard are unnecessarily complex. Composer plugins are capable of performing the necessary registration automatically, without requiring the user to manually set up (or alter) any `scripts` sections in `composer.json`. The main Magento repository uses `dealerdirect/phpcodesniffer-composer-installer` to register the coding standard; let's use the same here.

See https://github.com/magento/magento2/blob/11846a1a10539470f2fe1522030ff42d62daa562/composer.json#L110 / https://github.com/magento/magento2/pull/36791

Fixes #390

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -9,6 +9,7 @@
     "version": "40",
     "require": {
         "php": "~8.1.0 || ~8.2.0 || ~8.3.0 || ~8.4.0 || ~8.5.0",
+        "dealerdirect/phpcodesniffer-composer-installer": "^0.7.0 || ^1.0",
         "webonyx/graphql-php": "^15.0",
         "ext-simplexml": "*",
         "ext-dom": "*",
@@ -35,13 +36,9 @@
             "Magento2Framework\\": "Magento2Framework/"
         }
     },
-    "scripts": {
-        "post-install-cmd": "vendor/bin/phpcs --config-set installed_paths ../../..,../../magento/php-compatibility-fork/PHPCompatibility",
-        "post-update-cmd": "vendor/bin/phpcs --config-set installed_paths ../../..,../../magento/php-compatibility-fork/PHPCompatibility"
-    },
     "config": {
         "allow-plugins": {
-            "dealerdirect/phpcodesniffer-composer-installer": false
+            "dealerdirect/phpcodesniffer-composer-installer": true
         }
     }
 }
--- a/README.md
+++ b/README.md
@@ -10,20 +10,6 @@ To use within your Magento 2 project you can use:
 composer require --dev magento/magento-coding-standard
 ```
 
-Due to security, when installed this way the Magento standard for phpcs cannot be added automatically.
-You can achieve this by adding the following to your project's `composer.json`:
-
-```json
-"scripts": {
-    "post-install-cmd": [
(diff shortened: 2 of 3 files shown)


composer.json
```diff
@@ -9,6 +9,7 @@
     "version": "31",
     "require": {
         "php": ">=7.4",
+        "dealerdirect/phpcodesniffer-composer-installer": "^0.7.2 || ^1.0",
```
Den4ik: I believe that `"^1.0"` will be enough

Den4ik (commented):
Hi @fredden Thanks for your update.
Please review my comment

composer.json
fredden: There's nothing here that's *not* compatible with `^0.7.0`. I've specifically kept this because I expect other packages may take a while to upgrade / allow the new version. If there's something that requires `^0.7`, it'll work today, but not after merging if we change this to `^1.0` only. Listing both supported versions means that interoperability with other packages is maintained; Composer will pick the "best" version, which is typically the highest mutually-supported version. If we use `^1.0` only here, then we're forcing others to upgrade their packages in order for their coding standard to be used in conjunction with this standard.

Den4ik (dismissed):
✅ Looks good for me

ihor-sviziev (dismissed):

Den4ik (dismissed):

sinhaparul:
@magento import PR to magento-commerce/magento-coding-standard repository

m2-github-services:
@sinhaparul the Pull Request is successfully imported.

ihor-sviziev (dismissed):

ihor-sviziev (dismissed):

ihor-sviziev (dismissed):

ihor-sviziev (dismissed):

ihor-sviziev (dismissed):

Den4ik (approved):

PR #439, opened 2023-03-07
Allow more version numbers in `@deprecated` tag

fredden:
While reviewing the Static Test failures in https://github.com/magento/magento2/pull/36976 (which are out of scope for the intended change), I noticed that the message required for a `@deprecated` tag does not support the version numbers that Adobe use for Magento modules. For example, the `magento/module-backend` module is currently version `102.0.6`, but the regular expression in the coding standard does not allow version numbers with more than one digit per part. This pull request fixes this oversight.


ihor-sviziev (changes requested):
Could you please cover your change with automated test?

ihor-sviziev (dismissed):

PR #449, opened 2023-05-26
Auto-fix `Magento2.Whitespace.MultipleEmptyLines`

fredden:


Magento2/Tests/Whitespace/MultipleEmptyLinesUnitTest.inc.fixed
```diff
@@ -0,0 +1,24 @@
...
+
+        $someValue = 2;
+
+        $someValue = 3;
+
+        $someValue = 40;
+
+        $someValue = 'newlines and spaces and tabs';
+
+$someValue = 'no indentation';
```
ihor-sviziev, line 20: it doesn't look ok. are you sure it's fine?

fredden: @ihor-sviziev what do you expect to see here? This sniff deals with "multiple empty lines".

ihor-sviziev: I thought having spaces would be good. But in terms of this rule, it does make sense. Thx for the explanation

ihor-sviziev (approved):

PR #468, opened 2023-09-14
Removing dependency on symfony/polyfill, is no longer required now th…

### LINKED ISSUE #467: Individual polyfill packages should be used instead of symfony/polyfill
Currently, this package depends on `symfony/polyfill` instead of the individual `symfony/polyfill-*` packages.
This package installing the whole mono-repo is discouraged as it installs much more stuff than necessary.

The Symfony team (which I'm a member of) [recommends using the individual packages](https://github.com/symfony/polyfill/tree/1.x#usage) instead. If we were starting it today, we would not even publish the `symfony/polyfill` package on Packagist at all, keeping it a development-only mono-repo.

Currently, this even triggers a weird behavior in Composer. See https://github.com/composer/composer/issues/11641

hostep:
…at we dropped support for PHP 7.4

Fixes #467

As explained in the issue, the symfony/polyfill dependency got added in 2ea3c393903fcbc1db310d96e0aff4bba540c4a3 for the use of the php function `str_starts_with` which [was added in PHP 8.0](https://www.php.net/manual/en/function.str-starts-with.php), at that time, the coding-standards repository still supported PHP 7.4, so it made sense at the time to add a polyfill.
However, since 8d37ab7b8eb9c204abf2adcb2e637065595021c5 support for PHP 7.4 was dropped. So this polyfill package is no longer needed.

I'd say, merge this and tag a new version 33 of coding-standards, so people using Magento aren't being forced to install the entire `symfony/polyfill` package which comes with a whole lot of extra code that's not needed, maybe it will even cause a slight performance impact if all that code gets potentially loaded inside a Magento application.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -15,7 +15,6 @@
         "phpcompatibility/php-compatibility": "^9.3",
         "squizlabs/php_codesniffer": "^3.6.1",
         "rector/rector": "^0.17.12",
-        "symfony/polyfill": "^1.16",
         "phpcsstandards/phpcsutils": "^1.0.5"
     },
     "require-dev": {
--- a/composer.lock
+++ b/composer.lock
@@ -4,7 +4,7 @@
         "Read more about it at https://getcomposer.org/doc/01-basic-usage.md#installing-dependencies",
         "This file is @generated automatically"
     ],
-    "content-hash": "2898da982b9bfa7d7caa896d1a5d8944",
+    "content-hash": "1a1a72f8272e9cd2cb0dd520a56aeea2",
     "packages": [
         {
             "name": "dealerdirect/phpcodesniffer-composer-installer",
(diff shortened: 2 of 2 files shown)


fredden (approved):
Thanks for putting this together @hostep. I agree this can be safely removed. The coding standard should be installed as a development dependency though, so shouldn't cause any performance impact on production.

hostep:
@fredden: that's not exactly true, dependencies of dev-dependencies can influence production dependencies being installed.
See the output in https://github.com/composer/composer/issues/11641, by upgrading to v32 of coding-standards, it replaces the individual polyfill packages with the full one, and that one gets stored in the `composer.lock` file, so it gets installed in production as well.

I'm not a super fan of how dev dependencies work in composer, I've been experimenting with https://github.com/bamarni/composer-bin-plugin for over a year in some modules I've built ([example](https://github.com/baldwin-agency/magento2-module-image-cleanup/tree/v1-develop/vendor-bin)), for better seperation of dev & prod composer dependencies so they don't influence each other as much, and I quite like it. It doesn't work with everything, I believe phpunit and composer-normalize can't be used without problems this way, but most dev dependencies (phpcs, php-cs-fixer, phpstan, ...) can be used without issues like this.

sidolov:
@magento import pr to magento-commerce/magento-coding-standard repository

m2-github-services:
@sidolov the Pull Request is successfully imported.

sidolov:
@magento import pr to magento-commerce/magento-coding-standard repository

m2-github-services:
@sidolov the Pull Request is successfully imported.

fredden:
@sidolov please can we get a new release which includes this change?

hostep:
Got released in [v33](https://github.com/magento/magento-coding-standard/releases/tag/v33) today 🎉

PR #472, opened 2023-11-15
Rector rule added to escape unsafe output in phtml files

Jakhotiya:
## Summary
Often a codebase is handed over to you which does not follow all magento standards and phpcs starts flagging them. One such example is unsafe output in template files. For a large number of files, its cumbersome to refactor these files.
Rector rule can do this quickly. This pull requests addresses that.

## Checks done.
Added test cases for the rector rule. 
Ran it on actual template files

## Things not covered
This does not yet implement escapeJs and escapeUrl functions. Should be able to add it soon. This PR should also help people to get started with writing such rules.

### DIFF
--- a/Magento2/Rector/Tests/AddHtmlEscaperToOutput/Fixture/complex-phtml-code-mix.php.inc
+++ b/Magento2/Rector/Tests/AddHtmlEscaperToOutput/Fixture/complex-phtml-code-mix.php.inc
@@ -1,0 +1,27 @@
+<?php
+$items = $block->getAllItems();
+?>
+<div class="test">
+    <?php if (count($items) > 0): ?>
+        <div class="counter"><?= __('Total <span>%1</span> items found', count($items)) ?></div>
+        <?php foreach ($items as $item): ?>
+            <a href="<?= $block->getActionUrl($item); ?>"><?= $item->getName(); ?></a>
+        <?php endforeach; ?>
+    <?php else: ?>
+        <?= __('No items found.'); ?>
+    <?php endif; ?>
+</div>
+-----
+<?php
+$items = $block->getAllItems();
+?>
+<div class="test">
+    <?php if (count($items) > 0): ?>
+        <div class="counter"><?= __('Total <span>%1</span> items found', count($items)) ?></div>
+        <?php foreach ($items as $item): ?>
+            <a href="<?= $block->getActionUrl($item); ?>"><?= $escaper->escapeHtml($item->getName()); ?></a>
+        <?php endforeach; ?>
+    <?php else: ?>
+        <?= $escaper->escapeHtml(__('No items found.')); ?>
+    <?php endif; ?>
+</div>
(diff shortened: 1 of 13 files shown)


fredden:
I like the idea of auto-fixing this category, however I don't have confidence that this won't break things. Please can you add tests covering the following cases:

```php
<?php
print $var;
print($var);
print_r($var);
printf('%s', $var);
?>
<?= $var; ?>
<input value="<?= $var ?>">
<input value="<?php echo $var; ?>">
<?= /* @noEscape */ $var ?>
<a href="<?= $escaper->escapeUrl($var) ?>">
    <?= count($var) ?>
</a>
<span>PHP constant: <?= PHP_VERSION ?></span>
<?= 'hard-coded string' ?>
<?= "hard-coded string" ?>
<?= "String with interpolated $var" ?>
<span>Cast to "safe" type: <?= (int) $var ?></span>
<?= $block->getChildHtml() ?>
```

I can think of some more, but these should give better coverage and highlight some things that are likely to break. See `Magento2/Sniffs/Security/XssTemplateSniff.php` for what's considered safe / unsafe within this standard.

Jakhotiya:
Thank you Dan for feedback. Let me add those tests. I see your concern. Espcially when we do something like 

```
<?= $block->getChildHtml() ?>
```

OR 

```
<?php
$productCard = $block->getLayout()->getBlock();
$productCard->setProduct($product)
?>
<?= $productCard->toHtml() ?>
```

Predominatly things will break if we escape valid HTML itself. This espcially the case when Renderer blocks are used.
And looking at vendor/magento/module-catalog/view/frontend/templates/product/list.phtml I don't see an easy way out. 

A naive way is to ignore all methods with "Html" in them, Any suggestions on this @fredden ?

Jakhotiya:
@fredden I've escaped code only when echo statements are used
1. it's just a variable
2. when method name does not contain "Url" or "Html"
3. when @noEscape is not used
4. When __() contains string that does not contain html

Have tried to do it defensively. This does not intend to fix all escaping issues. Only where we are confident.

Risky case: 
```
<?php

$list = '';

foreach($block->getItems() as $item){
    $list .= '<li>'.$item->getName().'</li>';
}
?>
<ul><?= $list; ?></ul>
```

I've added a failing test on my local but haven't fixed it. 

The way I'm using this is by using dry-run to be safe

Magento2/Rector/Tests/AddHtmlEscaperToOutput/Fixture/complex-phtml-code-mix.php.inc
```diff
@@ -0,0 +1,27 @@
...
+</div>
+-----
+<?php
+$items = $block->getAllItems();
+?>
+<div class="test">
+    <?php if (count($items) > 0): ?>
+        <div class="counter"><?= __('Total <span>%1</span> items found', count($items)) ?></div>
+        <?php foreach ($items as $item): ?>
+            <a href="<?= $block->getActionUrl($item); ?>"><?= $escaper->escapeHtml($item->getName()); ?></a>
```
ihor-sviziev, line 22: shouldn't `$block->getActionUrl($item);` be wrapped to escapeUrl? or escapeHtmlAttr (because it's part of attribute)?

PR #476, opened 2023-12-14
Member variables and methods should not require DocBlock when properly typed

JacobBrownAustin:
https://wiki.corp.adobe.com/pages/viewpage.action?spaceKey=MC&title=Proposal+to+change+magento-coding-standard+to+not+require+DocBlock+when+parameters+defined


fredden:
@JacobBrownAustin thanks for raising this. Please can you add some tests to cover the changes being introduced here. I don't know why the automated test runs failed in GitHub Actions; do the tests run successfully on your system?

JacobBrownAustin:
Yeah; I'm working on updating tests now.

fredden:
> https://wiki.corp.adobe.com/pages/viewpage.action?spaceKey=MC&title=Proposal+to+change+magento-coding-standard+to+not+require+DocBlock+when+parameters+defined

This URL doesn't work for me. Is this an internal reference within Adobe? What does this page say?

JacobBrownAustin:
Yeah I'll update the details if this to include the info from that page.  Basically, I proposed that we shouldn't need superfluous DockBlock  for property or method if the type is already defined.  It's also solving a separate issue I found more recently that readonly properties fail even when they have proper DockBlock.

fredden:
There we're a lot of changes in PHP_CodeSniffer v3.8.0 related to `readonly`. It's possible that the bug you mention in passing here has already been solved. Perhaps you could open an issue here so that can be discussed/investigated.

fredden:
See also #406

Magento2/Sniffs/Annotation/MethodArgumentsSniff.php
```diff
@@ -617,6 +622,104 @@ public function process(File $phpcsFile, $stackPtr)
...
+     * @return bool
+     */
+    private function checkIfNamespaceContainsApi(File $phpcsFile) : bool
+    {
+        $namespaceStackPtr = $phpcsFile->findNext(T_NAMESPACE, 0);
+        if (!is_int($namespaceStackPtr)) {
+            return false;
+        }
+        $tokens = $phpcsFile->getTokens();
+        // phpcs:ignore Squiz.ControlStructures.ForLoopDeclaration.SpacingAfterOpen,Generic.CodeAnalysis.ForLoopWithTestFunctionCall.NotAllowed
```
fredden: Why not respect the coding standard?
```suggestion
```

Magento2/Sniffs/Annotation/MethodArgumentsSniff.php
```diff
@@ -617,6 +622,104 @@ public function process(File $phpcsFile, $stackPtr)
...
+        for (
+            // phpcs:ignore Squiz.ControlStructures.ForLoopDeclaration.SpacingAfterFirst
```
fredden: Why not respect the coding standard?
```suggestion
```

Magento2/Sniffs/Annotation/MethodArgumentsSniff.php
```diff
@@ -617,6 +622,104 @@ public function process(File $phpcsFile, $stackPtr)
...
+            $index = $namespaceStackPtr;
+            // phpcs:ignore Squiz.ControlStructures.ForLoopDeclaration.SpacingAfterSecond
```
fredden: Why not respect the coding standard?
```suggestion
```

Magento2/Sniffs/Annotation/MethodArgumentsSniff.php
```diff
@@ -617,6 +622,104 @@ public function process(File $phpcsFile, $stackPtr)
...
+            array_key_exists($index, $tokens) && 'T_SEMICOLON' !== $tokens[$index]['type'];
+            $index++
+            // phpcs:ignore Squiz.ControlStructures.ForLoopDeclaration.SpacingBeforeClose
```
fredden: Why not respect the coding standard?
```suggestion
```

Magento2/Sniffs/Annotation/MethodArgumentsSniff.php
JacobBrownAustin: For Generic.CodeAnalysis.ForLoopWithTestFunctionCall.NotAllowed: 
As you may know, since PHP 7.4, array_key_exists is no longer a PHP function.  It is compile to its own opcode, which has dramatically increased its performance.  Calling array_key_exists is O(1) since PHP arrays are implemented as hash tables which have O(1) CPU time complexity for lookups.
https://www.php.net/manual/en/migration74.other-changes.php
```A specialized VM opcode for the [array_key_exists()](https://www.php.net/manual/en/function.array-key-exists.php) function has been added, which improves performance of this function if it can be statically resolved. In namespaced code, this may require writing \array_key_exists() or explicitly importing the function```
Though, it looks like I messed up because I forgot to either use the `\` prefix or import the function.

For Squiz.ControlStructures.ForLoopDeclaration.SpacingAfterOpen:
Because the line that starts the for loop became over 120 characters long, I had to separate it into multiple lines.

Comment (JacobBrownAustin) [reply Magento2/Sniffs/Annotation/MethodArgumentsSniff.php]: I've resolved these in 17e120ff553a9b5123c3b11602cdb98491293fe6 by replacing array_key_exists with the null coalescing operator.

File: Magento2/Sniffs/Annotation/MethodArgumentsSniff.php (hunk 4 above)
Comment (JacobBrownAustin) [reply Magento2/Sniffs/Annotation/MethodArgumentsSniff.php]: I've resolved these in 17e120f by replacing array_key_exists with the null coalescing operator.

File: Magento2/Sniffs/Annotation/MethodArgumentsSniff.php (hunk 3 above)
Comment (JacobBrownAustin) [reply Magento2/Sniffs/Annotation/MethodArgumentsSniff.php]: I've resolved these in 17e120f by replacing array_key_exists with the null coalescing operator.

File: Magento2/Sniffs/Annotation/MethodArgumentsSniff.php (hunk 2 above)
Comment (JacobBrownAustin) [reply Magento2/Sniffs/Annotation/MethodArgumentsSniff.php]: I've resolved these in 17e120f by replacing array_key_exists with the null coalescing operator.

**Comment by JacobBrownAustin** [comment]
@fredden , I just noticed that this PR is still open.  Is there any changes that you want me to make?

File: Magento2/Tests/Annotation/MethodArgumentsUnitTest.inc [hunk 5]
```diff
@@ -41,13 +41,63 @@ public function invalidDocBlockShouldNotCauseFatalErrorInSniff(int $number): int
...
+}
+
+/**
+ * Short description.
+ *
+ * @param string $text
+ * @return string
+ */
+#[\ReturnTypeWillChange]
+public function methodWithAttributeAndValidDocblockWithoutTypes()
```
fredden, line 56: This method has no `$text` parameter. If this is an intentional test, please add a comment to show the same.

Magento2/Sniffs/Annotation/MethodArgumentsSniff.php
```diff
@@ -617,6 +622,96 @@ public function process(File $phpcsFile, $stackPtr)
...
+        $foundReturnType = !!$methodProperties['return_type'];
+        if (!$foundReturnType) {
+            $methodName = $phpcsFile->getDeclarationName($stackPtr);
+            // Note: __construct and __destruct can't have return types
+            if ('__construct' !== $methodName && '__destruct' !== $methodName) {
+                return true;
+            }
+        }
+        $complexity = $this->getMethodComplexity($phpcsFile, $stackPtr);
+        return $complexity > static::MAXIMUM_COMPLEXITY_ALLOWED_FOR_NO_DOCBLOCK;
```
fredden, line 655: Please add a test to cover this case - where a docblock is required due to high complexity score.

Magento2/Sniffs/Annotation/MethodArgumentsSniff.php
```diff
@@ -617,6 +622,96 @@ public function process(File $phpcsFile, $stackPtr)
...
+     * @param File $phpcsFile
+     * @param int $stackPtr
+     * @return bool
+     */
+    private function checkIfMethodNeedsDocBlock(File $phpcsFile, $stackPtr) : bool
+    {
+        if ($this->checkIfNamespaceContainsApi($phpcsFile)) {
+            // Note: API classes MUST have DocBlock because it uses them instead of reflection for types
+            return true;
+        }
```
fredden, line 637: Please add a test to cover this case - where a docblock is required due to a class being an API.

fredden (commented):
This looks good to me. It would be good to get some more test coverage.

Issue #485, opened 2024-08-29
Several include-path regex issues on both Windows and Linux


gwharton:
### Preconditions
1. Latest Magento Coding Standards Package. Reproduced on version 33.

### Steps to reproduce
The following creates a blank project, installs the necessary composer packages, creates some test files, and then patches phpcs to provide additional debugging output on the console for regex matching of include paths.
```
mkdir testproject
cd testproject
composer require magento/magento-coding-standard

tee test.php >/dev/null <<'EOF'
<?php
class Test extends ParentClassTest
{
    public function __construct() {
        parent::__construct();
    }
}
EOF

tee test.xml >/dev/null <<'EOF'
<?xml version="1.0" encoding="UTF-8" ?>
<root>
    <test>Hello</test>
</root>
EOF

tee test.less >/dev/null <<'EOF'
.selector {
    width: 0;
}
EOF

tee test.html >/dev/null <<'EOF'
<html>
    <head>
    </head>
    <body>
    </body>
</html>
EOF

tee test.phtml >/dev/null <<'EOF'
<p>Some Paragraph Text</p>
<?php
$test="Test";
echo $escaper->escapeHtml($test);
?>
EOF

mkdir -p anything/view/adminhtml/anything
mkdir -p anything/etc
mkdir -p etc/anything
mkdir -p etc/adminhtml
cp -f test.xml anything/view/adminhtml/anything/test.xml
cp -f test.xml anything/di.xml
cp -f test.xml anything/module.xml
cp -f test.xml anything/widget.xml
cp -f test.xml etc/config.xml
cp -f test.xml etc/config.anything.xml
cp -f test.xml etc/anything/config.xml
cp -f test.phtml anything/test.phtml
cp -f test.xml anything/etc/test.xml
cp -f test.xml etc/adminhtml/system.xml
cp -f test.xml etc/config.xml
cp -f test.xml etc/anything/config.xml
cp -f test.xml etc/config.xml
cp -f test.xml etc/config.anything.xml
composer require vaimo/composer-patches
composer config extra.patches-search patches
mkdir patches

tee patches/additional-debug-output.patch >/dev/null <<'EOF'
@package squizlabs/php_codesniffer
--- src/Files/File.original.php	2024-08-29 10:37:43.349529594 +0100
+++ src/Files/File.php	2024-08-29 10:39:12.000000000 +0100
@@ -319,6 +319,7 @@
      */
     public function process()
     {
[...]

PR #486, opened 2024-08-29
Resolve issues with regex include and exclude patterns in Magento 2 ruleset.xml

gwharton:
See https://github.com/magento/magento-coding-standard/issues/485 for a detailed description of the issue.

These fixes makes the following assumptions and these will need to be verified by someone who knows why the rules were introduced in the first place and has more knowledge about phpcs and Magento than I do.

**Magento2.Legacy.Layout**
The existing rule seems to target files such as

- anything/view/adminhtml/anything/.xml
- anything/view/frontend/anything/.xml

I don't think thats right. I believe it should actually be targetting

- anything/view/adminhtml/anything/anything.xml
- anything/view/frontend/anything/anything.xml

The updated rule targets any xml file in any subfolder of a view/adminhtml, view/frontend or view/base folder. I think this was the intended target.

**Magento2.Html.HtmlBinding**
I believe the intention with this rule is to target any phtml file anywhere in the tree. The existing rule, as written, will target only files with the following paths

- .phtml
- somewhere/.phtml
- somewhere/somewhere/.phtml

In addition to this, escaping the first forward slash, as in the previous example, breaks things even more under windows. The replacement rule removes the escaping of forward slash so it works under both Linux and Windows, and corrects the rule to target paths such as

- something.phtml
- somewhere/something.phtml
- somewhere/somewhere/something.phtml

**Magento2.Legacy.ClassReferencesInConfigurationFiles**
I believe the intention with this rule is to target any xml file present in any etc folder, anywhere, but exclude wsdl.xml, wsdl2.xml and wsi.xml. There is no need to escape the leading forward slash, and by doing so, it stops the rule working under windows. The replacement rule removes the escaping of the first slash so it works under both Linux and Windows.

**Magento2.Legacy.ModuleXML**
**Magento2.Legacy.DiConfig**
**Magento2.Legacy.WidgetXML**
I believe the intention with these rules is to target any file named module.xml, widget.xml or di.xml, in any subfolder of the tree, but not the root. The existing rules escape the first forward slash, which results in the rule failing when running under windows. The rule is updated to remove the escaping so it works under both Linux and Windows. 

**QUESTION**: Why were all these forward slashes escaped in the rules. Is there something/some system I am overlooking that requires it?

For more detailed analysis see issue #485 

Tested under Ubuntu 24.04 and Windows 11/Cygwin

### DIFF
--- a/Magento2/ruleset.xml
+++ b/Magento2/ruleset.xml
@@ -212,7 +212,7 @@
         <exclude-pattern>*\.xml$</exclude-pattern>
     </rule>
     <rule ref="Magento2.Html.HtmlBinding">
-        <include-pattern>*\/.phtml$</include-pattern>
+        <include-pattern>*\.phtml$</include-pattern>
         <severity>9</severity>
         <type>warning</type>
         <exclude-pattern>*\.xml$</exclude-pattern>
--- a/Magento2/ruleset.xml
+++ b/Magento2/ruleset.xml
@@ -165,10 +165,10 @@
         <exclude-pattern>*\.js$</exclude-pattern>
     </rule>
     <rule ref="Magento2.Legacy.ClassReferencesInConfigurationFiles">
-        <include-pattern>*\/etc/*.xml$</include-pattern>
-        <exclude-pattern>*\/etc/wsdl.xml$</exclude-pattern>
-        <exclude-pattern>*\/etc/wsdl2.xml$</exclude-pattern>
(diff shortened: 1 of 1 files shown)


gwharton:
Let me know if you would prefer these 4 commits squashed into 1 to keep the history clean.

fredden (approved):

ihor-sviziev (approved):

Issue #484, opened 2024-08-05
PHP 8.3 error mentioning php 8.1 as version


MaximGns:
### Preconditions

1.magento/product-enterprise-edition: "2.4.7-p1"
2. magento/magento-coding-standard: "33"
3. Previous coding standard version --> 31

```
PHP 8.3.9 (cli) (built: Jul  2 2024 14:10:14) (NTS)
Copyright (c) The PHP Group
Zend Engine v4.3.9, Copyright (c) Zend Technologies
    with Zend OPcache v8.3.9, Copyright (c), by Zend Technologies
```

```
FILE: ...ta/******.php
--------------------------------------------------------------------------------
FOUND 1 ERROR AFFECTING 1 LINE
--------------------------------------------------------------------------------
 9 | ERROR | Readonly classes are not supported in PHP 8.1 or earlier.
   |       | (PHPCompatibility.Classes.NewReadonlyClasses.Found)
--------------------------------------------------------------------------------

```

fredden:
This comes from the configuration for PHPCompatibility found in the ruleset for Magento2.

https://github.com/magento/magento-coding-standard/blob/v33/Magento2/ruleset.xml#L770
https://github.com/PHPCompatibility/PHPCompatibility/?tab=readme-ov-file#testversion-in-the-ruleset-versus-command-line

MaximGns:
@fredden How to resolve this issue when using GrumPHP?

fredden:
@MaximGns There are instructions on the PHPCompatibility repository for how to set `testVersion` via command line and/or in a custom ruleset. I've not tested with two custom rulesets both setting the value to see what comes out. You should be able to run `phpcs --standard=Magento2 --runtime-set testVersion 8.3 ...` to tell PHPCompatibility that you want to validate your code against PHP version 8.3 instead of the version(s) listed within the Magento2 ruleset.

MaximGns:
@fredden thanks for explenation, but why not support 8.3 out of the box as magento itself also supports? why only php 8.1/8.2...

PR #487, opened 2024-09-10
Update supported PHP versions to match Magento

fredden:
Version 2.4.7-p2 (which is the latest at time of writing) supports PHP versions 8.1.x and 8.2.x and 8.3.x. This change updates the PHPCompatibility configuration to scan for these versions.

https://github.com/magento/magento2/blob/2.4.7-p2/composer.json#L21
https://github.com/magento/magento2/blob/2.4.7-p1/composer.json#L21
https://github.com/magento/magento2/blob/2.4.7/composer.json#L21

This is related to https://github.com/magento/magento-coding-standard/issues/484 and https://github.com/magento/magento-coding-standard/issues/491

### DIFF
--- a/Magento2/ruleset.xml
+++ b/Magento2/ruleset.xml
@@ -767,7 +767,7 @@
     </rule>
 
     <!-- PHPCompatibility configuration. -->
-    <config name="testVersion" value="8.1-8.2"/>
+    <config name="testVersion" value="8.1-8.3"/>
     <rule ref="PHPCompatibility">
         <exclude name="PHPCompatibility.Miscellaneous.RemovedAlternativePHPTags.MaybeASPOpenTagFound" />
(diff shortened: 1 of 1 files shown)


ihor-sviziev (approved):

vacla:
Any update for when this would be merged/released as 2.4.7 is out for a bit now and has also support for PHP 8.3 ?

Zaahed (approved):

indykoning:
Since Magento 2.4.8 was released PHP 8.2 is no longer supported. And 8.4 has been added to the supported list.
With PHP 8.2 only getting security updates from PHP until the end of this year.

PRs like this must get reviewed and merged in to prevent packages like these from becoming completely obsolete and unused by the community.

PR #500, opened 2025-08-26
Fix Extra @param found in method annotation with PHP Attributes

danielfilipek-ug:
The getMethodArguments method in Magento Coding Standard was incorrectly counting PHP constructor arguments when PHP attributes (e.g., #[SerializedName('created_at')]) were present. 

The sniffer would stop counting parameters after encountering an attribute, resulting in incorrect code validations and false errors in the quality control system.

### DIFF
--- a/Magento2/Sniffs/Annotation/MethodArgumentsSniff.php
+++ b/Magento2/Sniffs/Annotation/MethodArgumentsSniff.php
@@ -106,6 +106,34 @@ class MethodArgumentsSniff implements Sniff
     }
 
     /**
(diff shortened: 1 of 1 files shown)


fredden:
This looks like a complicated sniff. I expect it can be simplified by using the 'owner' detail that PHP_CodeSniffer provides to find the open/close parenthesis.

These properties may help for someone who wants to explore this further:
https://github.com/PHPCSStandards/PHP_CodeSniffer/blob/ca606d9f60838b9a96ceb12dec7b935d5d64efce/src/Tokenizers/PHP.php#L2879-L2883

